PR #21, opened 2017-07-24
Cap overalPct at 100% to avoid wrong values

marc921:


checks/process.go
```diff
@@ -260,6 +260,10 @@ func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
 	// on a 2-CPU-core system would be reported as 50% instead of 100%.
 	overalPct := (deltaProc / deltaTime) * 100
 
+	if overalPct > 100 {
```
conorbranagan: small nit: can you add a comment of explanation. something simple like "Sometimes we get values that don't make sense, so we clamp to 100%"

PR #23, opened 2017-07-31
Changed the I/O data from counters to rates.

marc921:
As asked by conorbranagan, I changed the I/O data from counters (read from /proc/<pid>/io) to rates. This branch uses the marc.brun/countingIO branch in the gopsutil/ repository (it needs the IOCountersStat object in process.go and the fillFromIO function in process_linux.go).


checks/process.go
```diff
@@ -187,6 +188,23 @@ func formatCommand(fp *process.FilledProcess) *model.Command {
 	}
 }
 
+func formatIO(fp *process.FilledProcess, lastIO *process.IOCountersStat, before time.Time) *model.IOStat {
+	diff := time.Now().Unix() - before.Unix()
```
conorbranagan: you can safe this duplication by using the `calculateRate` helper function as we do elsewhere. then you can just put everything inline and have a nice small function.
```
return &model.IOStat{
    ReadRate: calculateRate(fp.IOState.ReadCount, lastIO,ReadCount, before),
    ...
}

marc921: OK done ! However I think real_time.go already has I/O rates in its messages : ContainerRbps and ContainerWbps. They read using util/docker/cgroup/ IO function, that seems to do something similar to process.go fillFromIO function.

conorbranagan: Those are for the container-level, we also want at the process-level. Basically you can just attach the result of `formatIO` to that message type as you did with `CollectorProc`

PR #27, opened 2017-07-31
[docker] Collect network statistics by container

conorbranagan:
* Add logic to DockerUtil for resolving containers to their interfaces.
* Collect network stats (bytes, packets) for each container's interface.
* Attach stats to the RealTime and CollectProc messages.


util/docker/docker_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/DataDog/datadog-process-agent/util"
+	"github.com/docker/docker/api/types"
+	dockernetwork "github.com/docker/docker/api/types/network"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFindDockerNetworks(t *testing.T) {
+	assert := assert.New(t)
+
+	hostProc := "/tmp/test-find-docker-networks/proc/"
```
shang-wang: If 2 tests are running at the same time, would `hostProc` directory be a problem?

conorbranagan: Yes this is definitely true. As far as I can tell, tests are not run in parallel by default so I think we're ok for now. Let me know if you've found otherwise on parallel tests.

PR #28, opened 2017-08-01
Add ECS metadata to CollectorProc payload

conorbranagan:
The bulk of the interesting work is in https://github.com/DataDog/datadog-agent/pull/420 and https://github.com/DataDog/agent-payload/pull/10. Both are still pending but appear to be passing so we'll wait until those are merged in.


conorbranagan:
This is ready to now as the other PRs are both merged

util/ecs/ecs.go
```diff
@@ -0,0 +1,46 @@
...
+		return globalECSUtil.getMetadata()
+	}
+	return nil
+}
+
+type ecsUtil struct{}
+
+func (e *ecsUtil) getMetadata() *agentpayload.ECSMetadataPayload {
+	payload, err := agentecs.GetPayload()
+	if err != nil && err.Error() != lastErr {
```
conorbranagan: Yeah we're logging and swallowing the error on purpose. I'll add a note on that

PR #31, opened 2017-08-02
Support mac/darwin

conorbranagan:
Much of the work is in https://github.com/DataDog/gopsutil/pull/10

On the Agent side we:

* Made the Docker check not error/warn on mac, it's just silently skipped. Getting Docker working on mac (with stats, at least) is a whole can of worms for later on.
* Added some additional checks for cases where Mac isn't supported yet (e.g. IOStats, Net stats)

Now it builds and runs!

### DIFF
--- a/checks/process.go
+++ b/checks/process.go
@@ -68,10 +68,7 @@ func (p *ProcessCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Mess
 	for _, fp := range fps {
 		pids = append(pids, fp.Pid)
 	}
-	containerByPID, err := docker.ContainersForPIDs(pids)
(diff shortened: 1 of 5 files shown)


checks/process.go
```diff
@@ -163,12 +163,16 @@ func formatCommand(fp *process.FilledProcess) *model.Command {
 		Root:   "",    // TODO
 		OnDisk: false, // TODO
 		Ppid:   fp.Ppid,
-		Pgroup: fp.Pgrp,
```
conorbranagan: Removing pgroup because (a) we don't make it visible anywhere, (b) it's not available on mac.

PR #38, opened 2017-08-08
Agent on by default in container w/ sane defaults.

conorbranagan:
* Agent is enabled with container checks by default.
* Path environment variables are set correctly in the container (should also simplify our `docker run` command for everyone!)
* The existing `DD_PROCESS_AGENT_ENABLED` flag and `process_agent_enabled` config will enable process-level collection. This allows us to eventually pull the containers stuff into a different Agent and have this Agent off again by default. And this also means that any existing customers will have the same behavior as they do now if upgrading.

@mikezvi Let me know your thoughts on having this configuration behavior for the next Agent release.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -16,6 +16,11 @@ import (
 	"github.com/go-ini/ini"
 )
 
+var (
+	processChecks   = []string{"process", "rtprocess"}
+	containerChecks = []string{"container", "rtcontainer"}
+)
+
 // AgentConfig is the global config for the process-agent. This information
 // is sourced from config files and the environment variables.
 type AgentConfig struct {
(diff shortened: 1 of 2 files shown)


mikezvi:
It would be nice to understand performance impact of the container checks, if we're not going to let people turn them off.  If it's as small as anticipated, then this all looks good, but I wouldn't want people to update to `latest` and have too much extra CPU overhead with no way back (or even if we do, want to do things deliberately).

Also if we're releasing this before GA let's make sure we're not spamming a log with 403s.

conorbranagan:
* Definitely agree on perf, that's on my list to check. If it's too much overhead we'll see if we can cut it down, otherwise we won't enable by default.
* 403s will be fixed in https://github.com/DataDog/dd-go/pull/3616

conorbranagan:
I'll also added that they can disable this Agent altogether by explicitly setting the flags to `false`.

PR #48, opened 2017-08-18
Collect context switch counts

conorbranagan:
Using the [data we already get from gopsutil](https://github.com/DataDog/gopsutil/blob/dd/process/process_linux.go#L643-L654)


conorbranagan:
Indeed, it was a real failure. Fixed now.

PR #58, opened 2017-09-12
added StartedAt flag for containers

shang-wang:
We need to collect start time for containers in addition to create time. Start time would be when a stopped container gets started, or equal to create time if it's a new one. 

This is done by looking at the stat of cgroup directory of the containers. The field "modified time" or "mtime" is used. It should be accurate because the cgroup directory is only created when the container started.


util/docker/docker.go
```diff
@@ -449,7 +450,22 @@ func (d *dockerUtil) containers() ([]*Container, error) {
 			container.Network = NullContainer.Network
 		}
 
+		// we get the stat for cgroup directory and use the mtime for that dir to determine the start time for the container
+		// this should work because the cgroup dir for the container would be created only when it's started
+		cgroupDir := cgroup.cgroupFilePath("cpuacct", "")
```
conorbranagan: can you push this into a method of `ContainerCgroup`?

Issue #52, opened 2017-08-29
Glide.yaml - Missing DataDog/datadog-agent repo


eosrei:
https://github.com/DataDog/datadog-process-agent/blob/42c9fd1ec3bd1a7fd38564c7eb0af1e60de9b423/glide.yaml#L3-L6
 
https://github.com/DataDog/datadog-agent does not exist:
![404](https://user-images.githubusercontent.com/195061/29848907-57abb760-8cd8-11e7-987b-6f18954e3f58.png)

conorbranagan:
Now that this agent is open source (https://github.com/DataDog/datadog-agent), the URL and building should work. Closing now but please re-open if you are still having trouble building it

eosrei:
Thanks!

PR #61, opened 2017-09-20
If scheme is nil url.Parse will throw an error

bkochendorfer:
Although datadog-process-agent supports proxy urls like the following:
* example.com
* http://example.com

The scheme was passed in blank resulting in an error in parsing a url
like `://example.com`

Please see https://play.golang.org/p/PUug1jdGSh for an example of the
error we are experiencing.

I wanted to add some tests but this project is not buildable. There is another open issue regarding this problem https://github.com/DataDog/datadog-process-agent/issues/52 . Once corrected I am happy to add some tests.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -526,10 +526,21 @@ func constructProxy(host, scheme string, port int, user, password string) *url.U
 	}
 
 	var path string
-	if userpass != nil {
-		path = fmt.Sprintf("%s://%s@%s:%v", scheme, userpass.String(), host, port)
+	// scheme is occasionally passed in as nil
+	if scheme != nil {
+		if userpass != nil {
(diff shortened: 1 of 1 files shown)


bkochendorfer:
Any feedback on this?

irabinovitch:
@conorbranagan @jeremy-lq Can you help review?

conorbranagan:
@bkochendorfer Sorry for the delay on getting back to you. I have done a fix with tests in https://github.com/DataDog/datadog-process-agent/pull/63 that I will merge once the tests are passing.

As for the agent not building, we are expecting to open source the `datadog-agent` repo later this week so you should be able to build/test things after that point.

Thanks for the report!

bkochendorfer:
No problem @conorbranagan! Closing in favor of your pull.

PR #62, opened 2017-09-22
[WIP] Create process-agent docker image

hkaj:
This image uses the latest build and very simple config. It runs the agent as a non-root user, which I'm not sure will work for all features but we'll see.


dockerfiles/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+COPY datadog.conf /etc/dd-agent/datadog.conf
+COPY dd-process-agent.ini /etc/dd-agent/dd-process-agent.ini
+
+RUN  chmod g+x /process-agent \
+  && mkdir /var/log/datadog/ \
+  && chmod -R g+w /var/log/datadog
+
+USER 1001
+
+CMD /process-agent -config /etc/dd-agent/dd-process-agent.ini -ddconfig /etc/dd-agent/datadog.conf
```
conorbranagan: the `dd-proces-agent.ini` config is legacy and is no longer needed. just setting `ddconfig` is enough

hkaj:
moving to datadog-agent repo

PR #64, opened 2017-10-02
agent info command

shang-wang:
Agent `--info` command is useful for us to debug and troubleshoot for customer issues. 

## How it works
It uses `expvar` package from golang to expose interval variables to an http endpoint. The info message is templated, then there are `updateXXX()` functions called in appropriate locations throughout agent to update those variables, and `publishXXX()` functions are responsible to publish the variables to `http://localhost:6062/debug/vars`. When `--info` is passed to agent binary, `Info()` function is called and the template would `Execute()` and pull the latest values of those variables published at `http://localhost:6062/debug/vars`.

Notes:
We reuse port `6062` which is also the profiling port, because `expvar` would publish to the port that the app is listening. This should have little conflict as it's rare for a profiling and info command are run at the same time.

The way how current container is detected is a little hacky, because there's no reliable way of knowing that in a straightforward way. Open to suggestions.


agent/info.go
```diff
@@ -0,0 +1,322 @@
...
+		case *model.CollectorContainerRealTime:
+			isProc = false
+			count += len(msg.Stats)
+		case *model.CollectorContainer:
+			isProc = false
+			count += len(msg.Containers)
+		case *model.CollectorRealTime:
+			isProc = true
+			count += len(msg.Stats)
+		case *model.CollectorProc:
```
conorbranagan: CollectorProc has both containers and processes

agent/info.go
```diff
@@ -0,0 +1,322 @@
...
+	return infoContainerCount
+}
+
+func updateProcContainerCount(msgs []model.MessageBody) {
+	var count int
+	// this is the flag to determine what type of collections are we dealing with
+	var isProc bool
+	for _, m := range msgs {
+		switch msg := m.(type) {
+		case *model.CollectorContainerRealTime:
```
conorbranagan: I think you don't need to worry about the stats messages for counts

agent/info.go
shang-wang: What about container only situation?

conorbranagan: Only `CollectorContainer`

PR #67, opened 2017-10-19
Use docker util from datadog-agent, sync interfaces.

conorbranagan:
In order to have consistent APIs for interacting with docker, work has
been moved into the `DataDog/datadog-agent` repo. This change points to
that repo, removes our local docker utilities, and syncs/updates any
relevant changes.


conorbranagan:
Heads up @xvello @hkaj

xvello:
Thanks for the heads up! 🎉 

Types should be stable now, last major change will be the DockerUtil lifecycle to allow for running even if started before the docker_daemon, as we did in agent5: https://github.com/DataDog/integrations-core/pull/722

I'll tag you on the PR.

PR #72, opened 2017-11-14
Add support for datadog.yaml, Agent 6 binary

conorbranagan:
* Adds support for the `datadog.yaml` file used in Agent 6. The default location is `/etc/datadog-agent/datadog.yaml` per the packaged installation.
* Use the Agent 6 binary if available to get the hostname.

Config might look like:

```yaml
api_key: XXXX
dd_process_url: https://process.datad0g.com
process_config:
  enabled: "true"
```

### DIFF
--- a/agent/main.go
+++ b/agent/main.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"gopkg.in/yaml.v2"
 	"net/http"
 	_ "net/http/pprof"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


gmmeyer:
This isn't a comment on the PR specifically. I think it's fine as it is. It's a broader comment about our config file. 

I feel like we should have some setting that prevents the need to set what will likely be nearly identical settings over and over again. It seems like we shouldn't have to set both:
```yaml
dd_url: https://app.datadoghq.com 
dd_process_url: https://process.datadoghq.com
```

conorbranagan:
@gmmeyer I agree in general. We don't even need the dd_process_url in the default config if we don't think people will be changing it (which seems likely). There's a similar structure for the logs URL which maybe could also be consolidated.

@masci Good point. I don't think this configuration will change all that much and I'm sure the important stuff (the top-level API key, ..) will not change on the Agent side. Hopefully long term (long, long term) this will just go into the mainline Agent

PR #81, opened 2017-11-30
[proc] bumped process-agent based on amazon linux cgroup root patch

shang-wang:
This is based on this work: https://github.com/DataDog/datadog-agent/pull/880.


conorbranagan:
Can this get merged to 5.20.x instead and then we can merge that back to master

shang-wang:
This one had a wrong base, close it.

PR #87, opened 2017-12-21
Default listeners for docker + ecs

sunhay:
The latest ECS Fargate changes broke the default behavior for docker container collection, as a `listener` must be configured in order to find running docker containers.

Here, we're setting the yaml config `listener` defaults to docker + ecs so that we can try and find containers no matter what.

Also, I noticed that the tagger was not being initialized and stopped if there was no yaml config provided. This fixes that.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -93,9 +93,13 @@ func NewDefaultAgentConfig() *AgentConfig {
 		// This is a hardcoded URL so parsing it should not fail
 		panic(err)
 	}
+
+	_, err = container.GetContainers()
+	canAccessContainers := err == nil
+
 	ac := &AgentConfig{
 		// We'll always run inside of a container.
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -93,9 +93,13 @@ func NewDefaultAgentConfig() *AgentConfig {
...
 		panic(err)
 	}
+
+	_, err = container.GetContainers()
+	canAccessContainers := err == nil
+
 	ac := &AgentConfig{
 		// We'll always run inside of a container.
-		Enabled:       container.IsAvailable(),
+		Enabled:       canAccessContainers,
```
sunhay: `container.IsAvailable()` actually just looks to see if a `listener` in the configuration was set. It makes more sense to try and connect to the container provider to determine if this should be enabled.

conorbranagan (approved):

PR #88, opened 2017-12-28
Fix datadog.process.*.host_count metrics

lachlancooper:
The container and process metrics were swapped around, which I noticed today on a dashboard graph.


sunhay:
Nice catch!

PR #90, opened 2018-01-04
Add Kubernetes service tags and remove ECS package

sunhay:
- Deprecating ECS metadata payload & removing the ECS package since all relevant metadata will be sourced from the [datadog-agent tagger](https://github.com/DataDog/datadog-agent/tree/master/pkg/tagger/).

- The tagger does not extract the kubernetes `kube_service` tags so we're now extracting that ourselves by constructing a container -> service tags mapping.

Tested this locally via minikube, here's an example container:
```
{
      "type": "Docker",
      "id": "51775c3829b48f3016c7225ccec0b171bac4b651fe17484baa63415c64984469",
      "cpuLimit": 100,
      "memoryLimit": 178257920,
      "state": 3,
      "created": 1515093949,
      "memRss": 6250496,
      "memCache": 13152256,
      "started": 1515093949,
      "tags": [
        "kube_namespace:kube-system",
        "kube_container_name:kubedns",
        "kube_deployment:kube-dns",
        "pod_name:kube-dns-86f6f55dd5-sdskb",
        "kube_replica_set:kube-dns-86f6f55dd5",
        "docker_image:gcr.io/google_containers/k8s-dns-kube-dns-amd64:1.14.5",
        "image_name:gcr.io/google_containers/k8s-dns-kube-dns-amd64",
        "image_tag:1.14.5",
        "container_name:k8s_kubedns_kube-dns-86f6f55dd5-sdskb_kube-system_7c77b2a5-dea5-11e7-bdb9-080027e5e101_10",
        "container_id:51775c3829b48f3016c7225ccec0b171bac4b651fe17484baa63415c64984469",
        "kube_service:kube-dns"
      ]
}
```


checks/container.go
```diff
@@ -114,11 +112,17 @@ func fmtContainers(
 		sys2, sys1 := ctr.CPU.SystemUsage, lastCtr.CPU.SystemUsage
 
 		// Retrieves metadata tags
-		tags, err := tagger.Tag(docker.ContainerIDToEntityName(ctr.ID), true)
+		entityID := docker.ContainerIDToEntityName(ctr.ID)
+		tags, err := tagger.Tag(entityID, true)
 		if err != nil {
+			log.Infof("Error retrieving tags for container: %s", err)
```
conorbranagan: nit: warn/error level maybe?

checks/container.go
```diff
@@ -114,11 +112,17 @@ func fmtContainers(
...
 			tags = []string{}
 		}
 
+		if _, isPresent := containerServiceTags[entityID]; isPresent {
```
conorbranagan: nit: go style is usually `_, ok := ...`
also you can do, `serviceTags, ok := ...` and then append `serviceTags...`

checks/container.go
```diff
@@ -97,6 +93,8 @@ func fmtContainers(
 		lastByID[c.ID] = c
 	}
 
+	containerServiceTags := kubernetes.GetContainerServiceTags()
```
conorbranagan: nit: generally I prefer names that make the key/val clear so I may not have to look at the function API. Something like `serviceTagsByID` might be nice.

util/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,99 @@
...
+
+	globalKubeUtil *agentkubelet.KubeUtil
+	lastKubeErr    string
+)
+
+// InitKubeUtil initializes a global kubeUtil used by later function calls.
+// We keep track of our own global kubeUtil even though the agentkubelet already does to prevent a noisy log
+func InitKubeUtil() error {
+	if ku, err := agentkubelet.GetKubeUtil(); err == nil {
+		globalKubeUtil = ku
```
conorbranagan: can we debug log the error here for support purposes (we log at INFO by default, so it won't show)

PR #91, opened 2018-01-05
Send no hostname to backend on Fargate instances

sunhay:
Starting by simply detecting whether we're in a fargate instance, and overwriting the hostname (even if it's set via the `DD_HOSTNAME` env variable) to be an empty string.

We may need to send something unique here for the backend key but I'll look into alternatives on the backend first.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -274,6 +274,8 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 		// Fargate tasks should have no concept of host names, so we're using the task ARN.
 		if taskMeta, err := ecsutil.GetTaskMetadata(); err == nil {
 			cfg.HostName = fmt.Sprintf("fargate_task:%s", taskMeta.TaskARN)
+		} else {
+			log.Errorf("Failed to retrieve Fargate task metadata: %s", err)
 		}
 	} else if hostname, err := getHostname(cfg.DDAgentPy, cfg.DDAgentBin, cfg.DDAgentPyEnv); err == nil {
 		cfg.HostName = hostname
--- a/config/config.go
+++ b/config/config.go
@@ -13,8 +13,9 @@ import (
 
 	"github.com/DataDog/datadog-agent/pkg/util/container"
 	"github.com/DataDog/datadog-agent/pkg/util/docker"
-	"github.com/DataDog/datadog-agent/pkg/util/ecs"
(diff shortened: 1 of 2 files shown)


conorbranagan:
We use this hostname for sharding at the intake so I think we'd end up having more unbalanced load if we just do this. On the backend we don't actually parse the message at intake (for perf reasons) so I think we'll need to define our unique key here in the Agent if we want to keep balanced sharding.

A unique fargate key should be: `task_id + container id` (although there might be other options here)

sunhay:
Curious to know what you think about these two options:

1. Set `HostName` to something like `"fargate_task:TASK_ARN"` 
   - `TASK_ARN` looks like `arn:aws:ecs:eu-west-1:172597598159:task/648ca535-cbe0-4de7-b102-28e50b81e888` per https://github.com/DataDog/integrations-core/pull/923/files#diff-19c1dcfb45d81707322676b116bdf3e2R3. We might want to hash this on the client first.
   -  In the backend we can handle hostnames with the `"fargate_task"` prefix as a special case and omit storing it as a hostname. 
   - We could also generalize the prefix to `"no_host"` if we wanted to.
2. Set `HostName` as empty, and send `TASK_ARN` through a new header like `X-Dd-Shardkey` which is only used for intake sharding.
   -  Then in the backend these wont we can handle these host-less payloads as a special case that is valid as it has a `X-Dd-Shardkey`

conorbranagan:
The first approach feels nicer to me because most of the pipeline works the same and if we decide to change sharding, etc we have that control all on the backend using a standard hostname field.

config/config.go
```diff
@@ -267,11 +269,15 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 		return nil, err
 	}
 
-	hostname, err := getHostname(cfg.DDAgentPy, cfg.DDAgentBin, cfg.DDAgentPyEnv)
-	if err != nil {
-		hostname = ""
+	cfg.HostName = ""
+	if ecsutil.IsFargateInstance() {
+		// Fargate tasks should have no concept of host names, so we're using the task ARN.
+		if taskMeta, err := ecsutil.GetTaskMetadata(); err == nil {
```
conorbranagan: I think we should log this error?

conorbranagan (approved):
One last thing but then go ahead and :shipit:

PR #94, opened 2018-01-11
Ansible config parsing fixes and pin glide dependencies

sunhay:
This PR adds some more robust configuration parsing for true/false values.
- Ansible generated `datadog.conf` files via https://github.com/DataDog/ansible-datadog captialize true/false values:  `process_agent_enabled: True`

It also updates `glide.yaml` to pin dependencies to current `glide.lock` versions so that our glide lockfile/yaml can remain synced.


trusche:
👍 Thanks for this!

PR #96, opened 2018-01-16
Adds windows support to process agent

derekwbrown:
Generate windows executable for windows process agent.


glide.yaml
```diff
@@ -1,13 +1,15 @@
 package: github.com/DataDog/datadog-process-agent
 import:
   - package: github.com/DataDog/datadog-agent
-    version: f1d63c176a934c8b8aefa31e2c7d938b0e50e348
+    vcs: git
+    version: db/merge_no_docker
```
sunhay: Does it still need to be pinned to this branch, or can we change it to the latest `master`?

glide.yaml
```diff
@@ -1,13 +1,15 @@
...
+    repo: git@github.com:DataDog/datadog-agent.git
   - package: github.com/DataDog/agent-payload
-    version: fb25fcedf155de94e762080a914adb5436d45b9c
+  - package: github.com/DataDog/gopsutil
+    vcs: git
+    version: db/add_windows_checks
```
sunhay: Can be pinned to the latest master :)

config/config.go
```diff
@@ -266,9 +266,9 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 	}
 
 	// (Re)configure the logging from our configuration
-	if err := NewLoggerLevel(cfg.LogLevel, cfg.LogFile); err != nil {
-		return nil, err
-	}
+	//if err := NewLoggerLevel(cfg.LogLevel, cfg.LogFile); err != nil {
```
sunhay: I'm not sure this is entirely correct, so the configuration is completely ignored?

checks/process_windows.go
```diff
@@ -0,0 +1,55 @@
...
+		UserPct:    calculatePct((t2.User-t1.User)*100, deltaSys, numCPU),
+		SystemPct:  calculatePct((t2.System-t1.System)*100, deltaSys, numCPU),
+		NumThreads: fp.NumThreads,
+		Cpus:       []*model.SingleCPUStat{},
+		Nice:       fp.Nice,
+		UserTime:   int64(t2.User),
+		SystemTime: int64(t2.System),
+	}
+}
+func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
```
sunhay: Since this is in both nix + windows, it might make sense to throw this into `process.go`?

agent/main_windows.go
```diff
@@ -0,0 +1,334 @@
+// +build windows
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	//	"log"
```
sunhay: nit: remove this

agent/main_windows.go
```diff
@@ -0,0 +1,334 @@
...
+var defaultConfdPath = "c:\\programdata\\datadog\\conf.d"
+
+var ddconfigPath string //= flag.String("ddconfig", "", "Path to the datadog.yaml configuration file")
+var ddconfdPath string  //= flag.String("ddconfd", "", "Path to the conf.d directory that contains all integration config files")
+var winopts struct {
+	installService   bool
+	uninstallService bool
+	startService     bool
+	stopService      bool
+}
```
conorbranagan: nit: can you wrap all of these in a single var()  block and clean up the comments with `flag` code

util/container/containers_nodocker.go
```diff
@@ -7,5 +7,5 @@ import "github.com/DataDog/datadog-agent/pkg/util/docker"
 // GetContainers is the unique method that returns all containers on the host (or in the task)
 // and that other agents can consume so that we don't have to convert all containers to the format.
 func GetContainers() ([]*docker.Container, error) {
-	return *docker.Container{}, nil
+	return make([]*docker.Container, 0), nil
```
conorbranagan: this can just be nil, nil without any allocation

derekwbrown: @conorbranagan I did it initially with `nil,nil`; however, there's other code that iterates the return (and crashes).  Needs to be empty (so length is zero) I think.

config/config.go
derekwbrown: I'll put it back, but IIRC I did that because it's logging to console no matter what, and on windows, when running as a service console == /dev/null

util/container/containers_nodocker.go
conorbranagan: Hm you should be able to iterate on a nil map but maybe there's something weird going on. Not a huge deal either way.

agent/main_windows.go
derekwbrown: Sure.  Something to think about, we use `github.com/spf13/cobra` on the metrics agent.  Might be nice to standardize.

PR #101, opened 2018-02-05
Split config options out into per-platform files, so that defaults can

derekwbrown:
be assigned on a per-platform basis.  First up, log file location


sunhay:
Looks like the build is failing because you need new lines at the end of each file.

PR #114, opened 2018-03-05
Remove kubelet build tag for windows

sunhay:
We don't support containers on windows so there's really no point in starting the kubelet tagger (via the `kubelet` build tag)

This cleans up some extra logging that I found in `process.log` (instead of `process-agent.log`)


sunhay:
@derekwbrown Would you be able to confirm that building without the `kubelet` tag causes no issues when running the process-agent?

PR #115, opened 2018-03-08
[config] Support proxy config for Agent 6

conorbranagan:
Pushes the configuration of transport to the config package so we can share the transport creator in Agent 6 with full support proxy configuration built in. See [util/common.go](https://github.com/DataDog/datadog-agent/blob/bc1f3ca4afde0a553098efa5c11a0f309b6b4156/pkg/util/common.go#L186) for the full function.

The potential breaking change here is changing the timeouts for Agent 6. We could override these outside of getting the transport but I'm not sure it's worth it. I assume if the transport works for core Agent 6 then it should be good for us. Thoughts?

### DIFF
--- a/agent/collector.go
+++ b/agent/collector.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"io/ioutil"
 	"math/rand"
-	"net"
 	"net/http"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


jamesgunja:
Are these proxy support present in docker container images too?

conorbranagan:
@jamesgunja Yes this should work in v6.1.1. You can either set the `proxy` value in the datadog.yaml file or a `PROXY` environment variable.

jamesgunja:
I tried PROXY values in environment, its still a issue. 

Error while processing transaction: error while sending transaction, rescheduling it: Post https://6-1-2-app.agent.datadoghq.com/api/v1/check_run?api_key=*************************: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)


$ docker run ... -e DD_APM_ENABLED=true -e DD_PROCESS_AGENT_ENABLED=true -e HTTP_PROXY='http://US_NAME:PASSWORD:URL' -e HTTPS_PROXY='http://US_NAME:PASSWORD:URL' datadog/agent:6.1.2

conorbranagan:
@jamesgunja Those errors are actually coming from the main infrastructure agent. If you open a ticket with support and include a [flare](https://help.datadoghq.com/hc/en-us/articles/204991415-Send-logs-and-configs-to-Datadog-via-flare-command) they will be able to help you sort out this issue.

PR #116, opened 2018-03-14
[proc] Process CPU percentage fix

sunhay:
Bugfix: We were dividing on nix systems when we should have been multiplying.

Also: Hardcode 6.1.0 as the windows version for the next release
Also: Bump the `datadog-agent` dependency

### DIFF
--- a/checks/process_windows.go
+++ b/checks/process_windows.go
@@ -33,3 +33,23 @@ func formatCPU(fp *process.FilledProcess, t2, t1, syst2, syst1 cpu.TimesStat) *m
 		SystemTime: int64(t2.System),
 	}
 }
+
+func calculatePct(deltaProc, deltaTime, numCPU float64) float32 {
+	if deltaTime == 0 {
+		return 0
+	}
+
+	// Calculates utilization split across all CPUs. A busy-loop process
+	// on a 2-CPU-core system would be reported as 50% instead of 100%.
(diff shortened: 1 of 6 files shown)


checks/process_windows.go
```diff
@@ -33,3 +33,23 @@ func formatCPU(fp *process.FilledProcess, t2, t1, syst2, syst1 cpu.TimesStat) *m
...
+
+	// Sometimes we get values that don't make sense, so we clamp to 100%
+	if overalPct > 100 {
+		overalPct = 100
+	}
+
+	// In order to emulate task mgr, we divide by number of CPUs.
+	// Task mgr displays percentage of available CPU (so a busy loop process
+	// on a 2 core CPU is 50%)
+	return float32(overalPct / numCPU)
```
sunhay: Not 100% if the Windows case is correct. Don't we want this to be out of `100*NumCPU`? We give the option to normalize out of 100 on the frontend...

conorbranagan: Yes I think they should be the same and this was an error but I don't have a windows box to test to confirm.

derekwbrown: After long discussion, I believe it should be
`if overalPct > (numCPU * 100) {
    overalPct = numCPU * 100
}
return overalPct` 
(don't do the division).  At least for now.

PR #122, opened 2018-03-30
Moises.botarro/adding process args blacklist feature

moisesbotarro:
@DataDog/burrito 

Added the blacklist process args feature on the agent


config/config.go
```diff
@@ -307,6 +333,20 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 	return cfg, nil
 }
 
+// Compile each word in the list into a regex pattern to match against the cmdline arguments
+func CompileStringsToRegex(words []string) []*regexp.Regexp {
+	compiledRegexps := make([]*regexp.Regexp, 0, len(words))
+	for _, word := range words {
+		pattern := `(-{1,2}` + word + `[^= ]*[ =])([^ \n]*)`
+		r, err := regexp.Compile(pattern)
```
sunhay: You should be able to use: `regexp.MustCompile`. I think it's better to fail fast on invalid regexps than silently ignoring them (especially if a user thinks their sensitive data is actually being blacklisted)

config/config_test.go
```diff
@@ -67,6 +67,46 @@ func TestBlacklist(t *testing.T) {
 	}
 }
 
+func TestArgsBlacklist(t *testing.T) {
```
sunhay: Can we also have a benchmark added for blacklisting by regex? I'm curious to know how these regexes will perform.

This resource should be able to help:
https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go

config/config_test.go
```diff
@@ -67,6 +67,46 @@ func TestBlacklist(t *testing.T) {
...
+
+	cases := []struct {
+		cmdline       []string
+		parsedCmdline []string
+	}{
+		{[]string{"agent", "-password", "1234"}, []string{"agent", "-password", "********"}},
+		{[]string{"agent", "--password", "1234"}, []string{"agent", "--password", "********"}},
+		{[]string{"agent", "-password=1234"}, []string{"agent", "-password=********"}},
+		{[]string{"agent", "--password=1234"}, []string{"agent", "--password=********"}},
+		{[]string{"spidly", "-debug_port=2043"}, []string{"spidly", "-debug_port=2043"}},
```
sunhay: Lets split these no-change cases into it's own test function, so its a bit clearer for which cases it should be modified, and which should not.

config/config_test.go
```diff
@@ -67,6 +67,46 @@ func TestBlacklist(t *testing.T) {
...
+	mergedRegexs := append(defaultRegexs, customRegexs...)
+	t.Log("default regexp", defaultRegexs)
+	t.Log("custom regexp", customRegexs)
+	t.Log("merged regexp", mergedRegexs)
+
+	cases := []struct {
+		cmdline       []string
+		parsedCmdline []string
+	}{
+		{[]string{"agent", "-password", "1234"}, []string{"agent", "-password", "********"}},
```
sunhay: How does this play for casing of the arguments?  e.g. `[]string{"agent", "-PASSWORD", "1234"}`

Might be worth adding a test case or two for that as well. If it's not handled, should it be?

config/config.go
```diff
@@ -37,29 +37,37 @@ var (
...
+	HostName             string
+	APIEndpoint          *url.URL
+	LogFile              string
+	LogLevel             string
+	QueueSize            int
+	Blacklist            []*regexp.Regexp
+	ArgsBlacklist        []*regexp.Regexp
+	DefaultArgsBlacklist []*regexp.Regexp
+	UseDefArgsBlacklist  bool
+	CustomArgsBlacklist  []*regexp.Regexp
```
conorbranagan: I would be in favor of encapsulating all the blacklisting behavior into a `DataScrubber` or `Blacklister` (and a separate module since this one is already rather large). Then your would have a simple interface like like `cfg.scrubber.ScrubCmdline(...)`.

As an aside I think using Sentry's "data scrubber" language is nicer than "blacklist" for something that will be customer visible

config/config_test.go
```diff
@@ -67,6 +67,97 @@ func TestBlacklist(t *testing.T) {
...
+func BenchmarkRegexMatching100000(b *testing.B)  { benchmarckRegexMatching(100000, b) }
+func BenchmarkRegexMatching1000000(b *testing.B) { benchmarckRegexMatching(1000000, b) }
+
+var avoidOptimization []string
+
+func benchmarckRegexMatching(nbProcesses int, b *testing.B) {
+	runningProcesses := make([][]string, nbProcesses)
+	foolCmdline := []string{"python ~/test/run.py --password=1234 -password 1234 -open_password=admin -consul_token 2345 -blocked_from_yaml=1234 &"}
+
+	for i := 0; i < nbProcesses; i++ {
```
sunhay: I think instead of trying to run this over 1000000 processes, it makes more sense to iterate over `b.N` so we can find out the average cost per regex match call.

config/config.go
```diff
@@ -307,6 +333,18 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 	return cfg, nil
 }
 
+// Compile each word in the list into a regex pattern to match against the cmdline arguments
+func CompileStringsToRegex(words []string) []*regexp.Regexp {
+	compiledRegexps := make([]*regexp.Regexp, 0, len(words))
+	for _, word := range words {
+		pattern := `((?i)-{1,2}` + word + `[^= ]*[ =])([^ \n]*)`
```
sunhay: This regex is starting to look pretty gnarly. I'm curious about the benchmarks for how this will perform and what kind of computational cost it'll add to the agent. Can you post benchmarks to the RFC?

config/config_test.go
moisesbotarro: I've just updated the tests wrapping this part into a iteration over b.N to get an average value for the time that the agent will pass doing the matching on the host. Do you think that it's better to just iterate through the matching function to get the avg per regex call?

config/config.go
moisesbotarro: The RFC is updated with the benchmark considering the matching for different numbers of processes running on the host. I can update it to consider just the avg per regex call if it's better

config/config_test.go
sunhay: I think that's find to get the average time matching on the host. I'd kill the 10000 and 100000 benchmarks since they're noise (I'd be surprised to find a host running that many processes at once.

You'll also want to add `b.ResetTimer()` after your setup (and just before the `b.N` loop so that it's not calculated in the benchmarks. (as noted in https://golang.org/pkg/testing/#hdr-Benchmarks)

moisesbotarro: Thanks, Sunny. Just updated the tests with ResetTimer and killed the biggest ones

config/config.go
moisesbotarro: The RFC is updated with the new benchmark values

config/config.go
```diff
@@ -304,6 +314,12 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 		cfg.Transport.Proxy = cfg.proxy
 	}
 
+	fmt.Println("DataScrubber")
```
conorbranagan: These shouldn't be committed to code

config/data_scrubber.go
```diff
@@ -0,0 +1,62 @@
...
+
+// Hide any cmdline argument value whose key matchs one of the patterns on the argsBlacklist vector
+func (ds *DataScrubber) ScrubCmdline(cmdline []string) []string {
+	if !ds.Enabled {
+		return cmdline
+	}
+
+	rawCmdline := strings.Join(cmdline, " ")
+	for _, pattern := range ds.SensitiveWords {
+		rawCmdline = pattern.ReplaceAllString(rawCmdline, `${key}${delimiter}********`)
```
conorbranagan: why are backticks vs quotes used for strings here?

config/data_scrubber.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return newDataScrubber
+}
+
+// Compile each word in the list into a regex pattern to match against the cmdline arguments
+func CompileStringsToRegex(words []string) []*regexp.Regexp {
+	compiledRegexps := make([]*regexp.Regexp, 0, len(words))
+	for _, word := range words {
+		pattern := `(?P<key>( |-)(?i)` + word + `)(?P<delimiter> +|=)(?P<value>[^\s]*)`
+		r := regexp.MustCompile(pattern)
```
conorbranagan: is it worth having better error handling than a hard panic if someone gives a word that doesn't compile as a regex? is that even possible?

config/data_scrubber_test.go
```diff
@@ -0,0 +1,143 @@
...
+		{[]string{"agent", "--PASSword", "1234"}, []string{"agent", "--PASSword", "********"}},
+		{[]string{"agent", "--PaSsWoRd=1234"}, []string{"agent", "--PaSsWoRd=********"}},
+		{[]string{"java -password      1234"}, []string{"java", "-password", "", "", "", "", "", "********"}},
+	}
+
+	scrubber := setupDataScrubber()
+	t.Log("scrub enabled: ", scrubber.Enabled)
+	t.Log("default regexp", scrubber.DefaultSensitiveWords)
+	t.Log("custom regexp", scrubber.CustomSensitiveWords)
+	t.Log("merged regexp", scrubber.SensitiveWords)
```
conorbranagan: I generally don't think we want logging in tests like this committed. If you need extra context for failures then it can go in the message arg of the `assert`

config/data_scrubber.go
```diff
@@ -0,0 +1,62 @@
...
+	if !ds.Enabled {
+		return cmdline
+	}
+
+	rawCmdline := strings.Join(cmdline, " ")
+	for _, pattern := range ds.SensitiveWords {
+		rawCmdline = pattern.ReplaceAllString(rawCmdline, `${key}${delimiter}********`)
+	}
+
+	return strings.Split(rawCmdline, " ")
```
conorbranagan: Are we 100% certain that data will remain the same when we do the join/split and that there is no change anything of the args in `cmdline` will have a space?

config/data_scrubber.go
```diff
@@ -0,0 +1,62 @@
+package config
+
+import (
+	"regexp"
+	"strings"
+)
+
+var (
+	defaultSenstiveWords = []string{"password", "passwd", "mysql_pwd", "access_token", "auth_token", "api_key", "apikey", "secret", "credentials", "stripetoken"}
```
conorbranagan: nit: can you break these up over a couple lines for better readability

config/data_scrubber.go
```diff
@@ -0,0 +1,62 @@
...
+	DefaultSensitiveWords []*regexp.Regexp
+	CustomSensitiveWords  []*regexp.Regexp
+}
+
+func NewDefaultDataScrubber() *DataScrubber {
+	newDataScrubber := &DataScrubber{
+		Enabled:               true,
+		SensitiveWords:        CompileStringsToRegex(defaultSenstiveWords),
+		DefaultSensitiveWords: CompileStringsToRegex(defaultSenstiveWords),
+		CustomSensitiveWords:  make([]*regexp.Regexp, 0, 0),
```
conorbranagan: Is there any advantage to having these all separated and duped (in case of the defaults)? Could we just have a single list of regexps that is extended with the custom ones?

config/data_scrubber.go
moisesbotarro: We can use Compile instead of mustCompile which sends an error when we can't parse correctly the expression. We can send the error to the NewAgentConfig function and treat it like the other errors or disable the DataScrubber by default (which can maybe confuse the user). Which option is better here?

conorbranagan: What about a warning when something can't compile but continue on without adding it to the list. That way an invalid regex won't block the agent running? That or just bubble up the error as you were saying.

config/data_scrubber.go
moisesbotarro: I used them separately to be able to enable/disable the default words but since we've changed the behavior of the DataScrubber, you're right that it doesn't make sense to maintain two separated lists. I'll merge them in a single one.

config/data_scrubber.go
moisesbotarro: I can envision 2 cases where the cmdline can change:

1. It hasn't been splitted before when collected by AllProcesses(). So we'll end up splitting it. By printing the processes cmdline on "process.go", we verified that it happened sometimes.
2. It was partially splitted. If we have ["run", "-p 1234"] we'll produce ["run", "-p", "1234"]

I can prevent this modification for the cmdlines that don't match any pattern if it'd be better. At worst, we only modify the structure of the cmdlines that have an arg to be hidden. What do you think?

conorbranagan: Ok. As long as things work as expected on the backend in either case, it should be fine.

config/data_scrubber.go
conorbranagan: Yeah, that's a good point. I think you want to filter our non-alphanumeric characters from the word set so it doesn't change the behavior of the regex.

moisesbotarro: I've done it! I've kept the test to verify it the regexp was well compiled for safe reasons even though it doesn't seem to have a problem since we only allow word characters.

PR #124, opened 2018-04-13
bugfix: DD_HOSTNAME was always overridden

jtwaleson:
DD_HOSTNAME is parsed in mergeEnv, but this was always overriden in the
fargate or hostname detection methods

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -117,6 +117,7 @@ func NewDefaultAgentConfig() *AgentConfig {
 		MaxProcFDs:    200,
 		ProcLimit:     100,
 		AllowRealTime: true,
+		HostName:      "",
 		Transport: &http.Transport{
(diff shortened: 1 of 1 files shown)


jtwaleson:
Great, thanks. Is the datadog process agent included in the normal v6 agent? I'm running the agent within a (docker) container next to the application. The agent is installed in a non-default location (under /home/user). I could not get the process list working unless I manually ran the process agent as an additional process.

conorbranagan:
@jtwaleson It is included in the normal docker agent. It does not need to be turned on via a flag in the config or an environment variable though. See our docs on: https://docs.datadoghq.com/graphing/infrastructure/process/#docker-process-collection. If this is still not working, please open a support ticket with a flare (feel free to mention me for escalation) and we'll take a look ASAP!

PR #127, opened 2018-04-20
[windows] Adds command line processing to post the arguments.

derekwbrown:
Does so by querying WMI for every process run.

Also moves `allProcesses` out of gopsutil on Windows.  Implementation
has diverged sufficiently that it shouldn't be lumped in with gopsutil.

Retrieving the command line arguments must be done via WMI, which could
have bad performance characteristics, especially if new processes are
added frequently.

By default, the process agent will now keep a table of the processes,
and update it every 15 checks (configurable), to see if command line
arguments have changed.  It will detect the command line arguments
immediately when a new process is detected.

The configuration option "windows_refresh_interval" will change the
above collection interval, with the value (-1) disabling argument
checking completely.

In addition, if listing command line arguments is desirable, but
processes are created frequently, then the configuration option
"windows_collect_skip_new_args" can be set to true.  In this case,
the arguments won't be queried upon process startup, but will be
refreshed every 15 (default) check runs.


checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+		return map[uint32]Win32_Process{}, fmt.Errorf("could not get Process")
+	}
+	results := make(map[uint32]Win32_Process)
+	for _, proc := range dst {
+		results[proc.ProcessID] = proc
+	}
+	return results, nil
+
+}
+func getAllProcesses() (map[int32]*process.FilledProcess, error) {
```
conorbranagan: nit: can you space out functions with an empty line to match style elsewhere?

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil {
+		log.Debugf("Failed to open process token %v", err)
+		return
+	}
+	defer t.Close()
+	tokenUser, err := t.GetTokenUser()
+
+	user, domain, _, err := tokenUser.User.Sid.LookupAccount("")
+	return domain + "\\" + user, err
+
```
conorbranagan: nit: can you not add extra lines either at top or bottom of function defs

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+func convert_windows_string(winput []uint16) string {
```
conorbranagan: convertWindowsString (I see this function was copied - but might as well fix it)

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	return retstring
+}
+
+func parseCmdLineArgs(cmdline string) (res []string) {
+
+	blocks := strings.Split(cmdline, " ")
+	findCloseQuote := false
+	donestring := false
+	var stringInProgress string
```
conorbranagan: should also use bytes.Buffer for this

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+				ReadBytes:  ioCounters.ReadTransferCount,
+				WriteBytes: ioCounters.WriteTransferCount,
+			},
+			Username: username,
+		}
+	}
+	return procs, nil
+}
+
+func get_username_for_process(h syscall.Handle) (name string, err error) {
```
conorbranagan: getUsernameForProcess

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	for i := 0; i < len(winput); i++ {
+		if winput[i] == 0 {
+			break
+		}
+		retstring += string(rune(winput[i]))
+	}
+	return retstring
+}
+
+func parseCmdLineArgs(cmdline string) (res []string) {
```
conorbranagan: this function should have a unit test on it as it's pretty complex

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+			if findCloseQuote {
+				donestring = true
+			} else {
+				findCloseQuote = true
+			}
+		} else if numquotes == 2 {
+			stringInProgress = strings.Trim(b, "\"")
+			donestring = true
+		} else {
+			log.Warnf("Unexpected qutoes in string, giving up")
```
conorbranagan: - s/qutoes/quotes
- can you include the full command line in this log for context, otherwise it's hard to know what this warning means or how you might fix it

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+				System:    stime,
+				Timestamp: time.Now().UnixNano(),
+			},
+
+			CreateTime:  ctime,
+			OpenFdCount: int32(handleCount),
+			//Name
+			// Status
+			// UIDS
+			// GIDs
```
conorbranagan: can you remove these if they aren't used, same for CWD below

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	if allProcsSnap == 0 {
+		return nil, syscall.GetLastError()
+	}
+	procs := make(map[int32]*process.FilledProcess)
+
+	defer w32.CloseHandle(allProcsSnap)
+	var pe32 w32.PROCESSENTRY32
+	pe32.DwSize = uint32(unsafe.Sizeof(pe32))
+
+	fromWmi, _ := getProcessMapFromWmi()
```
conorbranagan: why is this error getting ignored?

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	return nil
+}
+
+func getProcessMapFromWmi() (map[uint32]Win32_Process, error) {
+	var dst []Win32_Process
+	q := wmi.CreateQuery(&dst, "")
+	err := wmi.Query(q, &dst)
+	if err != nil {
+		return map[uint32]Win32_Process{}, err
```
conorbranagan: can you just return a nil map here?

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+	if len(dst) == 0 {
+		return map[uint32]Win32_Process{}, fmt.Errorf("could not get Process")
```
conorbranagan: this error message could be more explicit

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+
+		var pmemcounter process.PROCESS_MEMORY_COUNTERS
+		if err = getProcessMemoryInfo(procHandle, &pmemcounter); err != nil {
+			continue
+		}
+
+		// shell out to getprocessiocounters for io stats
+		var ioCounters IO_COUNTERS
+		if err = getProcessIoCounters(procHandle, &ioCounters); err != nil {
+			continue
```
conorbranagan: for all these `continue` cases I'm in favor of having some debug or trace-level logging. We went without this in the initial cut of the linux AllProcesses and it bit us when we had an issue in the wild where processes were skipped without any logging

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+		procs[int32(pid)] = &process.FilledProcess{
+			Pid:     int32(pid),
+			Ppid:    int32(ppid),
+			Cmdline: parsedargs,
+			CpuTime: cpu.TimesStat{
+				User:      utime,
+				System:    stime,
+				Timestamp: time.Now().UnixNano(),
+			},
+
```
conorbranagan: nit: extra line

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	tokenUser, err := t.GetTokenUser()
+
+	user, domain, _, err := tokenUser.User.Sid.LookupAccount("")
+	return domain + "\\" + user, err
+
+}
+
+func convert_windows_string(winput []uint16) string {
+	var retstring string
+	for i := 0; i < len(winput); i++ {
```
conorbranagan: any reason you're using a loop like this vs a `for .. := range`?

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,306 @@
...
+	defer t.Close()
+	tokenUser, err := t.GetTokenUser()
+
+	user, domain, _, err := tokenUser.User.Sid.LookupAccount("")
+	return domain + "\\" + user, err
+
+}
+
+func convert_windows_string(winput []uint16) string {
+	var retstring string
```
conorbranagan: should use a bytes.Buffer

config/yaml_config.go
```diff
@@ -57,6 +57,10 @@ type YamlAgentConfig struct {
 		DDAgentEnv []string `yaml:"dd_agent_env"`
 		// Overrides the submission endpoint URL from the default
 		ProcessDDURL string `yaml:"process_dd_url"`
+		// Sets windows process table refresh rate (in number of check runs)
+		WinProcessRefreshInterval int `yaml:"windows_refresh_interval"`
+		// Enables/disables getting process arguments immediately when a new process is discovered
+		WinSkipProcessNewArgs bool `yaml:"windows_collect_skip_new_args"`
```
conorbranagan: maybe just `windows_collect_new_args` and default to `false`?

config/config.go
```diff
@@ -153,6 +157,10 @@ func NewDefaultAgentConfig() *AgentConfig {
 
 		// DataScrubber to hide command line sensitive words
 		Scrubber: NewDefaultDataScrubber(),
+
+		// Windows process config
+		WindowsProcessRefreshInterval: 60, // number of check runs
```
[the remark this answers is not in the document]
derekwbrown: We can change it; I thought check runs were 5 seconds, so 60 would be 5 minutes.  By default, though, the process will already have the arguments on the first run, this is just the fallback for the case where the arguments somehow change (or if the user has specifically disabled the retrieving of the arguments when the new process appears)

config/yaml_config.go
```diff
@@ -57,6 +57,10 @@ type YamlAgentConfig struct {
 		DDAgentEnv []string `yaml:"dd_agent_env"`
 		// Overrides the submission endpoint URL from the default
 		ProcessDDURL string `yaml:"process_dd_url"`
+		// Sets windows process table refresh rate (in number of check runs)
+		WinProcessRefreshInterval int `yaml:"windows_refresh_interval"`
```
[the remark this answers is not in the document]
derekwbrown: we could; was going for consistent naming with the agent5 config file.

checks/allprocesses_windows.go
derekwbrown: yup

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,460 @@
...
+			// which will cause us to thrash WMI once per check, which we don't
+			// want to do.
+			continue
+		}
+		cp, ok := cachedProcesses[pid]
+		if !ok {
+			// wasn't already in the map.
+			cp = cachedProcess{}
+
+			if cfg.WindowsProcessRefreshInterval != -1 && cfg.WindowsProcessAddNew {
```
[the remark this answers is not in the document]
derekwbrown: The config file if the user disabled it.

checks/allprocesses_windows.go
```diff
@@ -0,0 +1,460 @@
...
+		if checkCount == 0 {
+			log.Infof("Windows process arg tracking enabled, will be refreshed every %v checks", cfg.WindowsProcessRefreshInterval)
+			if cfg.WindowsProcessAddNew {
+				log.Infof("Will collect new process args immediately")
+			} else {
+				log.Warnf("Will add process arguments only upon refresh")
+			}
+		}
+
+	} else {
```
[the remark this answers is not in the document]
derekwbrown: it's not else if; when checkCount ==0 I want to hit both cases, but the block of logs I only want on the first run.

config/config.go
```diff
@@ -275,6 +283,10 @@ func NewAgentConfig(agentIni *File, agentYaml *YamlAgentConfig) (*AgentConfig, e
 		cfg.ContainerBlacklist = agentIni.GetStrArrayDefault(ns, "container_blacklist", ",", cfg.ContainerBlacklist)
 		cfg.ContainerWhitelist = agentIni.GetStrArrayDefault(ns, "container_whitelist", ",", cfg.ContainerWhitelist)
 		cfg.ContainerCacheDuration = agentIni.GetDurationDefault(ns, "container_cache_duration", time.Second, 30*time.Second)
+
+		// windows args config
+		cfg.WindowsProcessRefreshInterval = agentIni.GetIntDefault(ns, "windows_refresh_interval", cfg.WindowsProcessRefreshInterval)
```
conorbranagan: Since we do `n % cfg.WindowsProcessRefreshInterval` we need to make sure this is not `0` to avoid a divide by zero error

PR #129, opened 2018-04-23
Support the log_to_console config option

sunhay:
Supporting the logger to output to stdout/console.


conorbranagan:
Could you match the behavior we have for `DD_LOGS_STDOUT` and give an empty file to trigger https://github.com/DataDog/datadog-process-agent/blob/master/config/log.go#L298?

conorbranagan:
Alternatively if you prefer this interface and the explicitness then we should switch DD_LOGS_STDOUT to use this flag as well just so there's one way to do it.

PR #130, opened 2018-04-24
Work-around for golint issue

sunhay:
golint is having a bad day and is breaking our CI: https://github.com/golang/lint/issues/397

I think we should leave this work-around unmerged until the issue above has some sort of response and/or fix.


sunhay:
Closing as this was fixed upstream.

PR #132, opened 2018-04-26
Changed scrub_args to *bool to better handle default value

moisesbotarro:
@DataDog/burrito 
Changed `scrub_args` field on yaml file to *bool and tagged it with "omitempty" to assure that default value is set to true even though the field isn't present in the yaml file.
Added some tests.

### DIFF
--- a/config/yaml_config.go
+++ b/config/yaml_config.go
@@ -42,6 +42,7 @@ type YamlAgentConfig struct {
 		// A list of regex patterns that will exclude a process if matched.
 		BlacklistPatterns []string `yaml:"blacklist_patterns"`
 		// Enable/Disable the DataScrubber to obfuscate process args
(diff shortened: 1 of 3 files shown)


config/yaml_config.go
```diff
@@ -42,7 +42,7 @@ type YamlAgentConfig struct {
 		// A list of regex patterns that will exclude a process if matched.
 		BlacklistPatterns []string `yaml:"blacklist_patterns"`
 		// Enable/Disable the DataScrubber to obfuscate process args
-		ScrubArgs bool `yaml:"scrub_args"`
+		ScrubArgs *bool `yaml:"scrub_args,omitempty"`
```
conorbranagan: can you add a note on why this is a bool pointer as I did below? just for future readers :)

PR #134, opened 2018-05-02
Added environment variables to control process args scrubbing

moisesbotarro:
@DataDog/burrito 

Added the possibility to set the data scrubber behavior using environment variables. This is useful when we don't configure the Agent using a conf or yaml file like when deploying it in kubernetes nodes.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -79,7 +79,6 @@ func TestOnlyEnvConfig(t *testing.T) {
 }
 
 func TestOnlyEnvConfigArgsScrubbingEnabled(t *testing.T) {
-	os.Setenv("DD_SCRUB_ARGS", "true")
 	os.Setenv("DD_CUSTOM_SENSITIVE_WORDS", "*password*,consul_token,*api_key")
 
 	agentConfig, _ := NewAgentConfig(nil, nil)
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -78,6 +78,54 @@ func TestOnlyEnvConfig(t *testing.T) {
...
+	agentConfig, _ := NewAgentConfig(nil, nil)
+	assert.Equal(t, true, agentConfig.Scrubber.Enabled)
+
+	cases := []struct {
+		cmdline       []string
+		parsedCmdline []string
+	}{
+		{[]string{"spidly", "--mypasswords=123,456", "consul_token", "1234", "--dd_api_key=1234"},
+			[]string{"spidly", "--mypasswords=********", "consul_token", "********", "--dd_api_key=********"}},
+	}
```
conorbranagan: nit: the formatting is a bit weird here. maybe line break around the outside open/close brackets?

config/config_test.go
```diff
@@ -78,6 +78,54 @@ func TestOnlyEnvConfig(t *testing.T) {
 	os.Setenv("DD_API_KEY", "")
 }
 
+func TestOnlyEnvConfigArgsScrubbingEnabled(t *testing.T) {
+	os.Setenv("DD_SCRUB_ARGS", "true")
```
conorbranagan: Can we verify that this defaults to true correctly as well without setting this explicitly to `true`?

moisesbotarro: Yep, it sounds a good idea. I've changed it

PR #140, opened 2018-05-23
BPF-based network connection collection

sunhay:
- Currently only enabled via the environment flag: `DD_CONNECTIONS_CHECK`.
- Also move to Circle CI 2.0 so we can use a Xenial-based docker image.

**Successful run**
```
$ sudo ./process-agent --check connections
Waiting 10 seconds to allow for active connections to transmit data
-----------------------------


Results for check connections
-----------------------------

{
  "hostName": "vagrant",
  "connections": [
    {
      "pid": 30203,
      "laddr": {
        "ip": "10.0.2.15",
        "port": 22
      },
      "raddr": {
        "ip": "10.0.2.2",
        "port": 54536
      },
      "bytesSent": 4912,
      "bytesRecieved": 36
    },
    {
      "pid": 30230,
      "laddr": {
        "ip": "10.0.2.15",
        "port": 47382
      },
      "raddr": {
        "ip": "91.189.88.23",
        "port": 80
      },
      "bytesSent": 240,
      "bytesRecieved": 61342907
    }
  ]
}
```

**Error when running without sudo:**
```
./process-agent --check connections
...
2018-05-30 19:34:08 ERROR (net.go:38) - failed to create network tracer: error while loading map "maps/tcp_stats_ipv4": operation not permitted
```


config/config.go
```diff
@@ -158,7 +158,7 @@ func NewDefaultAgentConfig() *AgentConfig {
 			"rtprocess":   2 * time.Second,
 			"container":   10 * time.Second,
 			"rtcontainer": 2 * time.Second,
-			"connections": 3 * 60 * time.Minute,
+			"connections": 10 * time.Minute,
```
conorbranagan: is the overhead low enough that we can run this every 10-30 seconds?

checks/net.go
```diff
@@ -30,43 +50,73 @@ func (c *ConnectionsCheck) Endpoint() string { return "/api/v1/collector" }
...
+				Port: int32(c.SPort),
 			},
 			Raddr: &model.Addr{
-				Ip:   c.Raddr.IP,
-				Port: int32(c.Raddr.Port),
+				Ip:   c.Dest,
+				Port: int32(c.DPort),
 			},
+			BytesSent:     int64(c.SendBytes), // TODO: Consider sending rate instead of cumulative byte counts
+			BytesRecieved: int64(c.RecvBytes),
```
conorbranagan: I think moving to rates might be worth doing from the get-go as it makes the backend processing all the more easier. we can technically support rates with metrics (we do this for one metric now) but many other metrics from this and the infra agent are sent as rates.

checks/net.go
```diff
@@ -3,22 +3,42 @@ package checks
...
 
 // Init initializes a ConnectionsCheck instance.
-func (c *ConnectionsCheck) Init(cfg *config.AgentConfig, sysInfo *model.SystemInfo) {}
+func (c *ConnectionsCheck) Init(cfg *config.AgentConfig, sysInfo *model.SystemInfo) {
+	var err error
+
+	// Checking whether the current kernel version is supported by the tracer
+	if c.supported, err = tracer.IsTracerSupportedByOS(); err != nil {
+		// err is always returned when false, so the above catches the !ok case as well
+		log.Errorf("network tracer unsupported by OS: %s", err)
```
conorbranagan: would this be run for folks on other systems? should it be a WARN?

agent/main_common.go
```diff
@@ -217,11 +217,16 @@ func debugCheckResults(cfg *config.AgentConfig, check string) error {
...
 	// Run the check once to prime the cache.
-	_, err := ch.Run(cfg, 0)
-	if err != nil {
+	if _, err := ch.Run(cfg, 0); err != nil {
 		return fmt.Errorf("collection error: %s", err)
 	}
-	time.Sleep(1 * time.Second)
+
+	if ch.Name() == checks.Connections.Name() {
+		fmt.Printf("Waiting 5 seconds to allow for active connections to transmit data\n")
```
conorbranagan: should we have such a wait in the main check as well? basically a warming period where it returns nothing for the first iterations of the check?

proto/agent.proto
```diff
@@ -294,14 +294,26 @@ message IOStat {
...
+enum ConnectionFamily {
+	v4 = 0;
+	v6 = 1;
+}
+
 message Connection {
 	int32 pid = 1;
-	int32 fd = 2;
+	// int32 fd = 2;         // Unused for now -- may be deprecated
 	int32 family = 3;
```
conorbranagan: change the type to ConnectionFamily/ConnectionType? (might need to be a new field, but it seems cleaner IMO).

config/config.go
sunhay: Ah good point - this was a typo. It should be `10 * time.Second`

agent/main_common.go
sunhay: I don't think it's needed while the agent is actually running since the BPF program will be loaded the entire time.

PR #141, opened 2018-06-05
Moises.botarro/creating cache for data scrubber

moisesbotarro:
@DataDog/burrito 

This PR adds a cache on DataScrubber to avoid using the regex matching for already know processes.
After each process check, we add one cycle on the cache and reaching the `cacheTTL` limit, it's reset. 

This must decrease the number of calls to `regexp.(*machine).tryBacktrack` that came up on the top of function calls while profiling the process-agent.

The following benchmark compares the approach without and with cache for a host with 1200 processes when every process stay on the cache and there aren't new process coming up.

|benchmark           | old ns/op |     new ns/op |     delta|
|---|---|---|---|
|BenchmarkCache-2|     37082806|      555633        |-98.50%|

A test to check the cache use has been implemented too.

### DIFF
--- a/config/data_scrubber.go
+++ b/config/data_scrubber.go
@@ -104,7 +104,7 @@ func compileStringsToRegex(words []string) []*regexp.Regexp {
 	return compiledRegexps
 }
 
-// createProcessKey return a unique identifier for a given process
+// createProcessKey returns an unique identifier for a given process
 func createProcessKey(p *process.FilledProcess) string {
 	var b bytes.Buffer
 	b.WriteString("p:")
--- a/config/data_scrubber.go
+++ b/config/data_scrubber.go
@@ -30,7 +30,7 @@ type DataScrubber struct {
 	SensitivePatterns []*regexp.Regexp
 	seenProcess       map[string]struct{}
 	cachedCmdlines    map[string][]string
-	cacheRuns         uint32
+	cacheCycles       uint32
 	cacheTTL          uint32
 }
 
@@ -42,7 +42,7 @@ func NewDefaultDataScrubber() *DataScrubber {
 		SensitivePatterns: compileStringsToRegex(defaultSensitiveWords),
 		seenProcess:       make(map[string]struct{}),
 		cachedCmdlines:    make(map[string][]string),
-		cacheRuns:         0,
+		cacheCycles:       0,
 		cacheTTL:          defaultCacheTTL,
 	}
 
@@ -119,7 +119,6 @@ func createProcessKey(p *process.FilledProcess) string {
 // process' cmdlines
 func (ds *DataScrubber) ScrubCmdlineWithCache(p *process.FilledProcess) []string {
 	pKey := createProcessKey(p)
-
 	if _, ok := ds.seenProcess[pKey]; !ok {
 		ds.seenProcess[pKey] = struct{}{}
 		ds.cachedCmdlines[pKey] = ds.ScrubCmdline(p.Cmdline)
@@ -128,14 +127,14 @@ func (ds *DataScrubber) ScrubCmdlineWithCache(p *process.FilledProcess) []string
 	return ds.cachedCmdlines[pKey]
 }
 
-// IncreaseCacheAge increases one cicle of cache memory age. If it reaches the
+// IncreaseCacheAge increases one cycle of cache memory age. If it reaches the
 // TTL, the cache is restarted
 func (ds *DataScrubber) IncreaseCacheAge() {
-	ds.cacheRuns++
-	if ds.cacheRuns == ds.cacheTTL {
+	ds.cacheCycles++
+	if ds.cacheCycles == ds.cacheTTL {
 		ds.seenProcess = make(map[string]struct{})
 		ds.cachedCmdlines = make(map[string][]string)
-		ds.cacheRuns = 0
+		ds.cacheCycles = 0
 	}
 }
 
--- a/config/data_scrubber.go
+++ b/config/data_scrubber.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
+	"github.com/DataDog/gopsutil/process"
 	log "github.com/cihub/seelog"
 )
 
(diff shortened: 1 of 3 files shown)


config/data_scrubber.go
```diff
@@ -90,6 +104,40 @@ func compileStringsToRegex(words []string) []*regexp.Regexp {
...
+	b.WriteString(strconv.Itoa(int(p.Pid)))
+	b.WriteString("|c:")
+	b.WriteString(strconv.Itoa(int(p.CreateTime)))
+
+	return b.String()
+}
+
+// ScrubCmdlineWithCache uses a cache memory to avoid scrubbing already known
+// process' cmdlines
+func (ds *DataScrubber) ScrubCmdlineWithCache(p *process.FilledProcess) []string {
```
sunhay: Call this function `ScrubProcessCommand` (or something like that) -- theres no need to expose that it is cached. You can make the other one unexported.

config/data_scrubber.go
```diff
@@ -17,11 +19,19 @@ var (
 		"secret", "credentials", "stripetoken"}
 )
 
+const (
+	defaultCacheTTL = 100
```
sunhay: 100 seems a little high. That's a cache life of 16 minutes. Can we make this lower? 25?

config/data_scrubber.go
```diff
@@ -17,11 +19,19 @@ var (
...
+)
+
 // DataScrubber allows the agent to blacklist cmdline arguments that match
 // a list of predefined and custom words
 type DataScrubber struct {
 	Enabled           bool
 	SensitivePatterns []*regexp.Regexp
+	seenProcess       map[string]struct{}
+	cachedCmdlines    map[string][]string
```
sunhay: You don't need `seenProcess` since `cachedCmdlines` has the same key structure.

config/data_scrubber.go
```diff
@@ -90,6 +104,40 @@ func compileStringsToRegex(words []string) []*regexp.Regexp {
...
+		ds.seenProcess[pKey] = struct{}{}
+		ds.cachedCmdlines[pKey] = ds.ScrubCmdline(p.Cmdline)
+	}
+
+	return ds.cachedCmdlines[pKey]
+}
+
+// IncreaseCacheAge increases one cycle of cache memory age. If it reaches the
+// TTL, the cache is restarted
+func (ds *DataScrubber) IncreaseCacheAge() {
```
sunhay: nit: Increase -> Increment

config/data_scrubber.go
moisesbotarro: I kept it now since I store just the scrubbed cmdlines in this map

config/data_scrubber_test.go
```diff
@@ -43,6 +46,157 @@ func setupDataScrubberWildCard(t *testing.T) *DataScrubber {
...
+			[]string{"spidly", "", "", "--befpass=********", "", "", "onebefpass", "", "", "********", "", "", "--befpassCustom=1234"}},
+
+		{[]string{"spidly", "--afterpass=2043", "afterpass_1", "1234", "--befafterpass_1=1234"},
+			[]string{"spidly", "--afterpass=********", "afterpass_1", "********", "--befafterpass_1=1234"}},
+		{[]string{"spidly --afterpass=2043 afterpass_1 1234 --befafterpass_1=1234"},
+			[]string{"spidly", "--afterpass=********", "afterpass_1", "********", "--befafterpass_1=1234"}},
+		{[]string{"spidly   --afterpass=2043   afterpass_1   1234   --befafterpass_1=1234"},
+			[]string{"spidly", "", "", "--afterpass=********", "", "", "afterpass_1", "", "", "********", "", "", "--befafterpass_1=1234"}},
+
+		{[]string{"spidly", "both", "1234", "-dd_both", "1234", "bothafter", "1234", "--dd_bothafter=1234"},
```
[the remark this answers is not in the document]
moisesbotarro: Yes, I did enable `gofmt` on the atomBeautify package. I think that it's using one more tabbing because i'm splitting the initialization of the struct in two lines

config/data_scrubber.go
```diff
@@ -90,11 +104,53 @@ func compileStringsToRegex(words []string) []*regexp.Regexp {
...
+}
+
+// ScrubProcessCommand uses a cache memory to avoid scrubbing already known
+// process' cmdlines
+func (ds *DataScrubber) ScrubProcessCommand(p *process.FilledProcess) []string {
+	pKey := createProcessKey(p)
+	if _, ok := ds.seenProcess[pKey]; !ok {
+		ds.seenProcess[pKey] = struct{}{}
+		scrubbed, changed := ds.scrubCmdline(p.Cmdline)
+		if changed {
```
sunhay: nit: 

```
if scrubbed, changed := ds.scrubCmdline(p.Cmdline); changed {
   ...
}
```

config/data_scrubber.go
```diff
@@ -107,9 +163,9 @@ func (ds *DataScrubber) ScrubCmdline(cmdline []string) []string {
 	}
 
 	if changed {
-		return strings.Split(rawCmdline, " ")
+		return strings.Split(rawCmdline, " "), changed
 	}
-	return cmdline
+	return cmdline, changed
```
sunhay: you can reduce this to a single return now since `changed` is being emitted 

```
scrubbed := nil
if changed {
  scrubbed = strings.Split(rawCmdline, " ")
}
return scrubbed, changed
```

config/data_scrubber_test.go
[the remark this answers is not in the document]
sunhay: I see what's happening. This is really not readable though. Can you fix this to be like:

```
return []testCase{
 {
   []string{"spidly", "--befpass=2043", "onebefpass", "1234", "--befpassCustom=1234"},
   []string{"spidly", "--befpass=********", "onebefpass", "********", "--befpassCustom=1234"},
 },
 ...
}
```

config/data_scrubber.go
moisesbotarro: I've changed to a single return but I keep returning the original cmdline if anything hasn't been scrubbed to avoid crashing the old tests that use the scrubbing feature without cache

config/data_scrubber.go
```diff
@@ -17,11 +19,19 @@ var (
 		"secret", "credentials", "stripetoken"}
 )
 
+const (
+	defaultCacheTTL = 25
```
conorbranagan: nit: this name is confusing to me because it's not really a "time" but the number of scrubs in a cache period. can we rename or add a comment (or both, if necessary)

PR #142, opened 2018-06-07
[proc] return 0 rate if previous value is 0

shang-wang:
Return 0 if previous value is 0 while calculating the rate.

This is an attempt to fix the problem of getting really large network tx/rx rate on containers. @DataDog/burrito


sunhay:
Can we get some unit tests for this change?

PR #144, opened 2018-06-22
Batch large sets of network connections in groups of configurable size

sunhay:
- Network check now batches connections messages in groups of configurable size (default: 100).
- Similar to batching for processes and containers.
- Renamed some configuration logging and fields to help clarify things in the codebase.

@DataDog/burrito


checks/net.go
```diff
@@ -150,3 +149,35 @@ func formatType(f tracer.ConnectionType) model.ConnectionType {
 		return -1
 	}
 }
+
+func batchConnections(cfg *config.AgentConfig, groupID int32, cxs []*model.Connection) []model.MessageBody {
+	batches := make([]model.MessageBody, 0)
```
shang-wang: Isn't it that `batches := make([]model.MessageBody, 0, groupSize)`? No big deal anyway.

PR #147, opened 2018-06-27
Support additional endpoints in the Agent

conorbranagan:
Both Agent 5 and 6 configuration support additional endpoints.
  - Agent 5 config actually uses all the api keys provided now.
  - Agent 6 has the same format as datadog.yaml but under the `process`
  namespace.
- When we POST the collector will run concurrent requests to limit any
impact on collection latency.

See test cases for example config

### DIFF
--- a/agent/collector.go
+++ b/agent/collector.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"math/rand"
@@ -95,7 +96,12 @@ func (l *Collector) runCheck(c checks.Check) {
 }
 
 func (l *Collector) run(exit chan bool) {
-	log.Infof("Starting process-agent for host=%s, endpoint=%s, enabled checks=%v", l.cfg.HostName, l.cfg.APIEndpoint, l.cfg.EnabledChecks)
+	eps := make([]string, 0, len(l.cfg.APIEndpoints))
+	for _, e := range l.cfg.APIEndpoints {
+		eps = append(eps, e.Endpoint.String())
+	}
+	log.Infof("Starting process-agent for host=%s, endpoints=%s, enabled checks=%v", l.cfg.HostName, eps, l.cfg.EnabledChecks)
+
 	go handleSignals(exit)
 	heartbeat := time.NewTicker(15 * time.Second)
 	queueSizeTicker := time.NewTicker(10 * time.Second)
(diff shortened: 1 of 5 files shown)


agent/collector.go
```diff
@@ -169,79 +175,138 @@ func (l *Collector) postMessage(endpoint string, m model.MessageBody) {
...
+		l.updateStatus(statuses)
+	}
+}
+
+func (l *Collector) updateStatus(statuses []*model.CollectorStatus) {
+	curEnabled := atomic.LoadInt64(&l.realTimeEnabled) == 1
+
+	// If any of the endpoints wants real-time we'll do that.
+	var nextEnable int
+	for _, s := range statuses {
```
sunhay: I'm a little confused by this. If we get two statuses back and the first is enabled, and the latter is  disabled, wont we set `nextEnable` to 0 causing RT mode to be disabled it?

Perhaps I'm missing something?

sunhay: If something is wrong, then maybe something like the following would work?

```
// If any of the endpoints wants real-time we'll do that.
shouldEnableRT := false
for _, s := range statuses {
  shouldEnableRT = shouldEnableRT || (s.ActiveClients > 0 && l.cfg.AllowRealTime)
}

if curEnabled && !shouldEnableRT {
  log.Info("Detected 0 clients, disabling real-time mode")
  atomic.StoreInt64(&l.realTimeEnabled, 0)
} else if !curEnabled && shouldEnableRT {
  log.Info("Detected active clients, enabling real-time mode")
  atomic.StoreInt64(&l.realTimeEnabled, 1)
}
```

agent/collector.go
```diff
@@ -169,79 +175,138 @@ func (l *Collector) postMessage(endpoint string, m model.MessageBody) {
...
+		log.Info("Detected active clients, enabling real-time mode")
+		atomic.StoreInt64(&l.realTimeEnabled, 1)
+	case -1:
+		log.Info("Detected 0 clients, disabling real-time mode")
+		atomic.StoreInt64(&l.realTimeEnabled, 0)
+	}
+
+	// We will pick the maximum interval given since generally this is
+	// only set if we're trying to limit load on the backend.
+	maxInterval := 0 * time.Second
```
sunhay: You could simplify this function by throwing this `maxInterval` calculation in the above loop? Feel free to ignore if you think it won't actually simplify things

agent/collector.go
conorbranagan: Once we see an enable we break out of the loop so it shouldn't care about the next disable. I think the tests are covering this case since it can be confusing. Your logic does look simpler and easier to understand so I'm happy to switch it something like that

sunhay: Ah, I see -- I missed the `break`!

Yeah, this took me a few minutes to grok, so might be worth simplifying

PR #151, opened 2018-07-09
[proc] added timeout to process-agent payload send

shang-wang:
Added timeout to http client and a context with timeout per request when agent sends process payloads to DataDog.  @DataDog/burrito

### DIFF
--- a/agent/collector.go
+++ b/agent/collector.go
@@ -271,10 +271,7 @@ func (l *Collector) postToAPI(endpoint config.APIEndpoint, checkPath string, bod
 	req.Header.Add("X-Dd-Hostname", l.cfg.HostName)
 	req.Header.Add("X-Dd-Processagentversion", Version)
 
-	// we create a context with timeout and attach it to the http request, to make sure that it doesn't get stuck on
-	// any circumstances
-	start := time.Now()
-	ctx, cancel := context.WithTimeout(context.Background(), HTTPTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), ReqCtxTimeout)
 	defer cancel()
 	req.WithContext(ctx)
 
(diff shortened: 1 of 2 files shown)


agent/collector.go
```diff
@@ -270,12 +271,23 @@ func (l *Collector) postToAPI(endpoint config.APIEndpoint, checkPath string, bod
 	req.Header.Add("X-Dd-Hostname", l.cfg.HostName)
 	req.Header.Add("X-Dd-Processagentversion", Version)
 
+	// we create a context with timeout and attach it to the http request, to make sure that it doesn't get stuck on
+	// any circumstances
```
conorbranagan: This comment seems unnecessary to me as this is implied by having a context timeout.

agent/collector.go
```diff
@@ -270,12 +271,23 @@ func (l *Collector) postToAPI(endpoint config.APIEndpoint, checkPath string, bod
...
+	start := time.Now()
+	ctx, cancel := context.WithTimeout(context.Background(), HTTPTimeout)
```
conorbranagan: I think we want the timeout to be longer than the http client timeout to make sure the http request can finish. thoughts?

agent/util.go
```diff
@@ -2,6 +2,12 @@ package main
 
 import (
 	"strings"
+	"time"
+)
+
+const (
+	// HTTPTimeout is the timeout in seconds for process-agent to send process payloads to DataDog
+	HTTPTimeout = 60 * time.Second
```
conorbranagan: 60s feels too long to me. any reason this value was picked?

agent/collector.go
```diff
@@ -270,12 +271,23 @@ func (l *Collector) postToAPI(endpoint config.APIEndpoint, checkPath string, bod
...
+	defer cancel()
+	req.WithContext(ctx)
+
 	resp, err := l.httpClient.Do(req)
 	if err != nil {
 		if isHTTPTimeout(err) {
 			responses <- errResponse("Timeout detected on %s, %s", url, err)
 		} else {
-			responses <- errResponse("Error submitting payload to %s: %s", url, err)
+			if time.Since(start) >= HTTPTimeout {
```
conorbranagan: I'd rather just leave the error handling as before for simplicity. it will say timeout either way (from my testing the error message with a context cancel was: `req err: Get https://google.com: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers`

shang-wang: Really? If I set context timeout shorter than client timeout I got `Get http://127.0.0.1:3001: context deadline exceeded`.

shang-wang: But yea, we could just keep the original checks.

sunhay:
You'll probably want to cherry-pick this into https://github.com/DataDog/datadog-process-agent/tree/6.3.x, for the fix release

PR #158, opened 2018-07-25
[proc] made endpoints and api keys match length instead of returning an error

shang-wang:
Added logics to match the length of endpoints and api keys instead of returning an error. @DataDog/burrito

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -163,24 +163,45 @@ func TestGetHostname(t *testing.T) {
 }
 
 func TestDDAgentMultiAPIKeys(t *testing.T) {
+	// if no endpoint is given but api_keys are there, match the first api_key
+	// with the default endpoint
 	assert := assert.New(t)
 	ddAgentConf, _ := ini.Load([]byte("[Main]\n\napi_key=foo,bar "))
 	configFile := &File{instance: ddAgentConf, Path: "whatever"}
 	agentConfig, err := NewAgentConfig(configFile, nil)
-	assert.Error(err)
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -163,24 +163,45 @@ func TestGetHostname(t *testing.T) {
...
 	agentConfig, err = NewAgentConfig(configFile, nil)
+	assert.NoError(err)
+	assert.Equal(2, len(agentConfig.APIEndpoints))
 	assert.Equal("foo", agentConfig.APIEndpoints[0].APIKey)
 	assert.Equal("process.datadoghq.com", agentConfig.APIEndpoints[0].Endpoint.Hostname())
 	assert.Equal("bar", agentConfig.APIEndpoints[1].APIKey)
 	assert.Equal("process.datadoghq.eu", agentConfig.APIEndpoints[1].Endpoint.Hostname())
+
+	// if endpoint count is greater than api_key count, drop additional endpoints
+	ddAgentConf, _ = ini.Load([]byte(strings.Join([]string{
```
conorbranagan: Can we add a test for the reverse case? (More api keys than endpoints)

shang-wang: Well, the first test case in this test function is already doing that. If you don't provide any endpoints there's actually a default endpoint, so it's 2 api_keys and 1 endpoint.

shang-wang: But I should add some comment.

shang-wang: Oh wait, there's already comment :)

PR #163, opened 2018-08-06
Support for the new containers backends, kubelet support

xvello:
Rebase the process-agent on top of https://github.com/DataDog/datadog-agent/pull/2114 and https://github.com/DataDog/datadog-agent/pull/2081, and use the new interface. Most changes are detailed on the upstream PR. In a nutshell:

- All sources now provide a common `Collector` interface
- A new `kubelet` source works for docker, containerd and crio
- Detection is handled by the `Detector` class, with a retry mechanism to support running on the host and starting after the control plane
- The responsibility of keeping previous values and caching the container list between RT runs is shifted to the process-agent. It's all in the `util/containers.go` file
- Clean up dead code

## To be done: process-agent

- The `DD_PROCESS_AGENT_CONTAINER_SOURCE` allows to force the container source. Need to add it as configuration file field
- Right now, container collection is disabled if the process-agent does not find a docker socket. That needs to be replaced by a check on `GetContainers` returning a `collectors.ErrPermaFail` error (`ErrNothingYet` should make the agent keep retrying later). This means I cannot test the autodetection of the kubelet source, as the docker one will take priority
- Logs from `datadog-agent` packages don't show up anymore, added a hack in `main.go` for testing, needs to be replaced by better solution

## To be done: kubelet source

- For now, the `health` value is the readiness probe, not the liveness probe (not exposed in the podlist anyway)
- Need to decide how to display the container name, as we cannot get the real `container_name` tag. Shall we add a `display_container_name` tag, that we can format to our liking? Same for ecs where the raw `container_name` we get from the metadata is not great


checks/container.go
```diff
@@ -40,27 +40,27 @@ func (c *ContainerCheck) Endpoint() string { return "/api/v1/container" }
...
+	ctrList, err := util.GetContainers()
 	if err != nil {
 		return nil, err
 	}
 
 	// End check early if this is our first run.
-	if c.lastContainers == nil {
-		c.lastContainers = containers
+	if c.lastRates == nil {
+		c.lastRates = util.KeepContainerRateMetrics(ctrList)
```
conorbranagan: personal preference I think I'd call this `ExtractContainerRateMetric` which seems clearer to me that it will return something. `Keep` could imply some side-effect with no return.

util/containers_linux.go
```diff
@@ -0,0 +1,75 @@
...
+	"time"
+
+	"github.com/DataDog/datadog-agent/pkg/util/cache"
+	"github.com/DataDog/datadog-agent/pkg/util/containers"
+	"github.com/DataDog/datadog-agent/pkg/util/containers/collectors"
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+var detector *collectors.Detector
+var containerCacheDuration = 10 * time.Second
```
conorbranagan: nit: can you wrap these vars in a single var?

util/containers_linux.go
```diff
@@ -0,0 +1,75 @@
...
+	if !hit {
+		containers, err := l.List()
+		if err != nil {
+			return nil, err
+		}
+		cache.Cache.Set(cacheKey, containers, containerCacheDuration)
+		log.Infof("Got %d containers from source %s", len(containers), name)
+		return containers, nil
+	}
+	return nil, errors.New("")
```
conorbranagan: I could imagine some code change that has us hitting this empty error. should we just drop the `if !hit {` and follow the normal path? or just an `else`?

config/config.go
```diff
@@ -482,7 +484,9 @@ func mergeEnvironmentVariables(c *AgentConfig) *AgentConfig {
 		durationS, _ := strconv.Atoi(v)
 		c.ContainerCacheDuration = time.Duration(durationS) * time.Second
 	}
-
+	if v := os.Getenv("DD_PROCESS_AGENT_CONTAINER_SOURCE"); v != "" {
```
conorbranagan: Is there some configuration that can be shared with the main Agent? I've been trying to cut down on the number of config that's specifically for the process agent where possible.

xvello:
I merged the `datadog-agent` PR and tested the build, merging to master.

`display_container_name` works OK on the app, and nicely overrides the default docker `container_name` for docker containers. I'll prepare a PR to remove the synthetic `container_name` from the kubelet collector, as it's not useful anymore. We'll need to update the lock on `datadog-agent` before RC1.

PR #167, opened 2018-08-27
Collect network connections remotely via UDS

sunhay:
The process-agent can now communicate with a remote `network-tracer` via unix sockets.

### DIFF
--- a/agent/main_common.go
+++ b/agent/main_common.go
@@ -231,7 +231,7 @@ func printResults(cfg *config.AgentConfig, ch checks.Check) error {
 		return fmt.Errorf("collection error: %s", err)
 	}
 
-	if ch.Name() == checks.Connections.Name() {
+	if cfg.EnableLocalNetworkTracer && ch.Name() == checks.Connections.Name() {
 		fmt.Printf("Waiting 5 seconds to allow for active connections to transmit data\n")
 		time.Sleep(5 * time.Second)
 	} else {
--- a/config/config.go
+++ b/config/config.go
@@ -27,6 +27,8 @@ var (
 	// This mirrors the configuration for the infrastructure agent.
 	defaultProxyPort = 3128
 
+	defaultNetworkTracerSocketPath = "/var/run/datadog/nettracer.sock"
+
 	processChecks   = []string{"process", "rtprocess"}
 	containerChecks = []string{"container", "rtcontainer"}
 
@@ -76,6 +78,10 @@ type AgentConfig struct {
 	StatsdHost    string
 	StatsdPort    int
 
+	// Network collection configuration
+	EnableLocalNetworkTracer bool
+	NetworkTracerSocketPath  string
+
 	// Check config
 	EnabledChecks  []string
 	CheckIntervals map[string]time.Duration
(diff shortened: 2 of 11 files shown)


util/network_linux.go
```diff
@@ -0,0 +1,136 @@
+// +build linux
+
+package util
```
conorbranagan: I might be a stickler here (so feel free to ignore this one!) but this might make sense in a different package like `net`. `util` is notoriously generic, especially if this is extended. Again not a huge deal if it's left as-is

util/network_linux.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return status == retry.OK || !logged
+}
+
+func newNetworkTracer() *RemoteNetTracerUtil {
+	return &RemoteNetTracerUtil{
+		socketPath: globalSocketPath,
+		httpClient: http.Client{
+			Timeout: 5 * time.Second,
+			Transport: &http.Transport{ // TODO: Consider reducing these timeouts, since we're connecting to a unix socket
```
conorbranagan: The timeouts from our statsd UDS implementation are much more aggressive and might be a good one to check out: https://github.com/DataDog/datadog-go/blob/master/statsd/uds.go#L13

util/network_linux.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// GetConnections returns a set of active network connections, retrieved from the network tracer service
+func (r *RemoteNetTracerUtil) GetConnections() ([]tracer.ConnectionStats, error) {
+	// Otherwise, get it remotely (via unix socket), and parse from JSON
+	resp, err := r.httpClient.Get(connectionsURL)
+	if err != nil {
+		return nil, err
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("conn request failed: socket %s, url: %s, status code: %d", r.socketPath, connectionsURL, resp.StatusCode)
```
conorbranagan: would having the response body be useful for the error as well or will that be logged elsewhere?

config/config.go
```diff
@@ -114,6 +122,21 @@ const (
 	maxMessageBatch = 100
 )
 
+// DefaultTransport provides a http transport configuration with sane default timeouts
+func DefaultTransport() *http.Transport {
```
conorbranagan: `NewDefaultTransport`

agent/main_common.go
```diff
@@ -231,7 +231,7 @@ func printResults(cfg *config.AgentConfig, ch checks.Check) error {
 		return fmt.Errorf("collection error: %s", err)
 	}
 
-	if ch.Name() == checks.Connections.Name() {
+	if cfg.EnableLocalNetworkTracer && ch.Name() == checks.Connections.Name() {
```
conorbranagan: Can we have this work for a remote tracer too? Not worth it?

util/network_linux.go
sunhay: Agreed -- I originally did it because we have `util.containers`. Moved.

util/network_linux.go
sunhay: I dont think so since we control the server and know there isn't any useful data in the body :)

PR #170, opened 2018-08-29
Unused code + documentation

sunhay:
Couple of small changes. Added some documentation via comments for `DD_PROCESS_AGENT_CONTAINER_SOURCE` and a possible issue with container auto-detection on initialization.

Also removed `initMetadataProviders`, which is no longer needed as `docker.GetDockerUtil()` is already called down the stack when we call `collectors.NewDetector("").

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -484,9 +484,13 @@ func mergeEnvironmentVariables(c *AgentConfig) *AgentConfig {
 		durationS, _ := strconv.Atoi(v)
 		c.ContainerCacheDuration = time.Duration(durationS) * time.Second
 	}
+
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -122,6 +121,8 @@ func NewDefaultAgentConfig() *AgentConfig {
 		panic(err)
 	}
 
+	// Note: This only considers container sources that are already setup. It's possible that container sources may
+	//       need a few minutes to be ready.
```
sunhay: Since this is not trivial to fix, I've added a note about the caveat, and we can tackle it in the future.

PR #176, opened 2018-09-05
[container] protect against nil stats on containers from ecs fargate

truthbk:
Because we need to monitor for containers in other modes other than `NORMAL`, this (https://github.com/DataDog/datadog-agent/pull/2274) doesn't look like the way to go, unfortunately. So let's just be resilient on the process-agent side.

Hopefully the default values for the container model in case the rates are not computable/available are fine for the paused, CNI containers. Pushing this to try to unblock the `6.5.0 rc.2` release.

---

Also switch back to storing the previous container from just the rate metrics. This will simplfiy the interface for setting defaults on nil containers and make it easy to apply it broadly.

The trade-off is maybe more memory but we've had this in all versions up to this point so I'm not too worried about it.

Includes some simple tests that would panic before the code change.

### DIFF
--- a/checks/container.go
+++ b/checks/container.go
@@ -101,9 +101,19 @@ func fmtContainers(
 			lastCtr = util.NullContainerRates
 		}
 
+		// Just in case the container is found, but refs are nil
+		if lastCtr.CPU == nil {
+			lastCtr.CPU = util.NullContainerRates.CPU
+		}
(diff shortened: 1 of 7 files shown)


conorbranagan:
I'm going to update that this to work with real-time and simplify a bit + add a test. Should have something shortly.

PR #194, opened 2018-10-05
Move network-tracer agent to process-agent repo

sunhay:
Merging the network-tracer agent into this repository, for simpler deployment & development processes.

### DIFF
--- a/cmd/network-tracer/main.go
+++ b/cmd/network-tracer/main.go
@@ -78,7 +78,7 @@ func main() {
 	}
 
 	nettracer, err := CreateNetworkTracer(cfg)
-	if err != nil && strings.HasPrefix(err.Error(), TracerUnsupportedError.Error()) {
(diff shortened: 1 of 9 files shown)


cmd/network-tracer/main.go
```diff
@@ -0,0 +1,153 @@
...
+func handleSignals(exit chan bool) {
+	sigIn := make(chan os.Signal, 100)
+	signal.Notify(sigIn)
+	// unix only in all likelihood;  but we don't care.
+	for sig := range sigIn {
+		switch sig {
+		case syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL, syscall.SIGQUIT:
+			log.Criticalf("Caught signal '%s'; terminating.", sig)
+			close(exit)
+		default:
```
sfluor: I think it would be nice to have the same behavior as the process agent by logging even in the default case (for debugging purposes): https://github.com/DataDog/datadog-process-agent/blob/master/cmd/agent/main_nodocker.go#L24

PR #201, opened 2018-10-30
Prevent uint underflows when calculating rates and percentages

kevinconaway:
This PR addresses the scenario where some metrics were being reported as impossibly large values.

The cause here was unsigned integer underflow as when we calculate the "rate" for a given metric, we calculate `(current - previous) / elapsed time in seconds`.  In our case,  `current` and `previous` are usually unsigned integers so if `previous > current`, the subtraction operation will overflow.

This usually only occurs in corner cases where the container metrics were not able to be collected for some reason, possibly because the container is shutting down.

### DIFF
--- a/checks/container.go
+++ b/checks/container.go
@@ -159,9 +159,14 @@ func calculateCtrPct(cur, prev, sys2, sys1 uint64, numCPU int, before time.Time)
 		return 0
 	}
 
+	// Prevent uint underflows
+	if prev > cur || sys1 > sys2 {
+		return 0
+	}
+
 	// If we have system usage values then we need to calculate against those.
 	// XXX: Right now this only applies to ECS collection
-	if sys1 > 0 && sys2 > 0 {
+	if sys1 > 0 && sys2 > 0 && sys2 != sys1 {
 		cpuDelta := float32(cur - prev)
 		sysDelta := float32(sys2 - sys1)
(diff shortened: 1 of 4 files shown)


conorbranagan (approved):
LGTM, thanks

checks/container.go
```diff
@@ -159,9 +159,14 @@ func calculateCtrPct(cur, prev, sys2, sys1 uint64, numCPU int, before time.Time)
 		return 0
 	}
 
+	// Prevent uint underflows
+	if prev > cur || sys1 > sys2 {
```
leeavital: Should we be returning a negative percent here instead of zero?

leeavital: Spoke with Kevin offline, and I misunderstood what this was supposed to catch. 

Could this be more clearly written as: 

```
if cur == 0 || sys2 == 0 {
   return 0
}
```

Since zero is a value we're treating specially?

kevinconaway: For this case yes.  Given that the agent is more difficult to upgrade in the field, I'd prefer to be a bit defensive here though and guard against underflows in the case that the current values are non-zero and smaller than the previous ones which could occur in some future unforeseen circumstance.

sunhay (approved):
LGTM. Thanks for the increased test coverage 📈

PR #202, opened 2018-11-05
Ensure that process.IOStat is non-nil

kevinconaway:


checks/process.go
```diff
@@ -181,7 +181,7 @@ func formatIO(fp *process.FilledProcess, lastIO *process.IOCountersStat, before
 
 	diff := time.Now().Unix() - before.Unix()
 	if before.IsZero() || diff <= 0 {
-		return nil
```
leeavital: Should we change the signature to return a non-pointer struct if we never return nil?

sunhay (approved):

checks/process.go
kevinconaway: Right or wrong, the current approach is consistent with `formatMemory` and `formatCommand`.

Whats the benefit on returning a non pointer?

leeavital: The only benefit is that the compiler will force us to return a struct with zero-ed out fields instead of nil. Since we never return nil (and none of the callers handle nil returns) it feels off to allow it to be returned in the function signature.

However, going by https://stackoverflow.com/questions/23542989/pointers-vs-values-in-parameters-and-return-values it looks like pointers are recommended for "big structs" which IOStat is.

PR #207, opened 2018-11-13
Filter out connection stats for TCP connections attached to dead processes

leeavital:


leeavital:
There are a couple of other approaches I can think of:

- cache the os.Stat calls so we only check for PID liveness every few minutes per pid
- track some state in the BPF module that ages out PIDs that haven't been active for a while (similar to what we do for UDP)

cmd/network-tracer/tracer.go
```diff
@@ -91,3 +93,34 @@ func (nt *NetworkTracer) Close() {
...
+
+	deadProcs := 0
+	for _, conn := range conns.Conns {
+		if conn.Type == tracer.TCP && isDeadPID(conn.Pid) {
+			deadProcs++
+		} else {
+			filteredConnections = append(filteredConnections, conn)
+		}
+	}
+	log.Infof("there were %d dead connections in connection capture")
```
kevinconaway: Should this be at the debug level? Apart from startup, it doesn't look like we log at info anywhere else here.

leeavital:
Closing in favor of https://github.com/DataDog/tcptracer-bpf/pull/28

PR #208, opened 2018-11-15
properly compute location of programdata rather than hard-code

derekwbrown:
Title says it all

### DIFF
--- a/config/config_windows.go
+++ b/config/config_windows.go
@@ -2,7 +2,14 @@
 
 package config
 
-const (
+import (
+	"path/filepath"
+
+	"github.com/DataDog/datadog-agent/pkg/util/executable"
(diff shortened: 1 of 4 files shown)


config/config_windows.go
```diff
@@ -12,3 +19,15 @@ const (
 	// Agent 6
 	defaultDDAgentBin = "c:\\Program Files\\Datadog\\Datadog Agent\\embedded\\agent.exe"
 )
+
+func init() {
+	pd, err := winutil.GetProgramDataDir()
+	if err == nil {
```
sunhay: Just a style nit.

```go
if pd, err := winutil.GetProgramDataDir(); err == nil {
  ...
}
```

And the same below as well

PR #209, opened 2018-11-15
[proc] add container to LAddr and RAddr for connections

shang-wang:
Added container object for source and destination Addr so that they would be filled later with resolved container. It will not be a full container object but most likely just some required fields filled out only. @DataDog/burrito


proto/agent.proto
```diff
@@ -325,10 +325,18 @@ message Connection {
 	uint64 totalBytesReceived = 14;
 }
 
+// ContainerMetadata only holds enough information to identify a container in connection data
+message ContainerMetadata {
+	string id = 1;
+	uint32 key = 2;
```
leeavital: we should document what key and byteKey are in relation to a container

proto/agent.proto
```diff
@@ -325,10 +325,18 @@ message Connection {
...
+	bytes byteKey = 3;
```
sunhay: Should we include a tags array as well?

shang-wang:
I'm going to close this and add a slightly bigger PR to track more changes.

PR #210, opened 2018-11-16
Use DataDog/datadog-agent/pkg/config

sfluor:
Migrate to the main agent config package, this also includes some code splitting

Detail on why we can't load the ini inside the`ddconfig.Config` instance:

If we use BindEnv on variable `a` for instance and use `SetDefault` on it we have 2 cases:
- we either load the viper config after the ini config which means we will have to check if the variable has not been set already by the ini config (for each variable)
- we load it before the ini config, but then we can’t know If we can override one property since there is no way to distinguish properties set by the YAML conf// Env variable (we allow to use both ini and yaml on the process agent) or by `SetDefault`


config/merge_config.go
```diff
@@ -0,0 +1,393 @@
...
+
+	if dc.IsSet(keyStripProcessArguments) {
+		agentConf.Scrubber.StripAllArguments = dc.GetBool(keyStripProcessArguments)
+	}
+
+	if dc.IsSet(keyQueueSize) {
+		agentConf.QueueSize = dc.GetInt(keyQueueSize)
+	}
+
+	if dc.IsSet(keyMaxProcFDs) {
```
sunhay: There's a lot of duplication here, can we create a helper function to reduce a lot of it?

```go
func ifSet(dc ddconfig.Config, key string, f func(s string)) {
  if dc.IsSet(key) {
    f(s)
  }
}

... 

ifSet(dc, keyMaxProcFDs, func(key string) {
   agentConf.MaxProcFDs = dc.GetInt(key)
})
```

sfluor: I added helpers

config/merge_config.go
```diff
@@ -0,0 +1,377 @@
...
+	"github.com/go-ini/ini"
+)
+
+func mergeCheckIntervalIfSet(dc ddconfig.Config, agentConf *AgentConfig, key, mapKey string) {
+	if dc.IsSet(key) {
+		agentConf.CheckIntervals[mapKey] = time.Duration(dc.GetInt(key)) * time.Second
+	}
+}
+
+func mergeString(dc ddconfig.Config, key string, defaultVal string) string {
```
leeavital: can we call these get{String/Int/Bool}OrDefault -- elsewhere in the code `merge` implies that the parameters are being mutated

config/merge_config.go
```diff
@@ -0,0 +1,377 @@
...
+
+	return nil
+}
+
+// mergeEnvironmentVariables applies overrides from environment variables to the process agent configuration
+func mergeEnvironmentVariables(dc ddconfig.Config, c *AgentConfig) {
+	var err error
+
+	// Support API_KEY and DD_API_KEY but prefer DD_API_KEY.
+	if v := os.Getenv("API_KEY"); v != "" {
```
leeavital: I think this is redundant as https://github.com/DataDog/datadog-process-agent/pull/210/files#diff-c05fe2f63536332ce41f97d262a780ddR64 will also read from the environment

config/merge_config.go
```diff
@@ -0,0 +1,377 @@
...
+		log.Info("overriding API key from env API_KEY value")
+		mergeAPIKeys(v, c)
+	}
+
+	// Support LOG_LEVEL and DD_LOG_LEVEL but prefer DD_LOG_LEVEL
+	if v := os.Getenv("LOG_LEVEL"); v != "" {
```
leeavital: same comment as line 191

config/merge_config.go
```diff
@@ -0,0 +1,377 @@
...
+
+	agentConf.DisableTCPTracing, _ = isAffirmative(dc.GetString(keyNetworkDisableTCPTracing))
+	agentConf.DisableUDPTracing, _ = isAffirmative(dc.GetString(keyNetworkDisableUDPTracing))
+	agentConf.DisableIPv6Tracing, _ = isAffirmative(dc.GetString(keyNetworkDisableIPV6Tracing))
+
+	return nil
+}
+
+// mergeEnvironmentVariables applies overrides from environment variables to the process agent configuration
+func mergeEnvironmentVariables(dc ddconfig.Config, c *AgentConfig) {
```
leeavital: nit: if we can not avoid reading API_KEy and LOG_LEVEL in this function, maybe call this mergeEnvironmentOnlyConfig as this handles all the things that are *not* configured in configuration files. `mergeEnvironmentVariables` feels wrong because `mergeConfig` will read environment variables if it finds them.

config/merge_config.go
sfluor: Unfortunately it will only read the `DD_API_KEY` env variable and not `API_KEY` (and we support both)

config/merge_config.go
sfluor: same as above with `DD_LOG_LEVEL` and `LOG_LEVEL`

config/merge_config.go
```diff
@@ -0,0 +1,378 @@
...
+	if dc.IsSet(key) {
+		agentConf.CheckIntervals[mapKey] = time.Duration(dc.GetInt(key)) * time.Second
+	}
+}
+
+func getString(dc ddconfig.Config, key string, defaultVal string) string {
+	if dc.IsSet(key) {
+		return dc.GetString(key)
+	}
+
```
sunhay: nit: but you dont need these empty lines (also for the functions below too)

cmd/agent/main_common.go
```diff
@@ -97,18 +98,16 @@ func runAgent(exit chan bool) {
...
 
-	agentConf, err := config.NewIfExists(opts.ddConfigPath)
+	agentConf, err := config.NewReaderIfExists(opts.ddConfigPath)
 	if err != nil {
 		log.Criticalf("Error reading dd-agent config: %s", err)
 		os.Exit(1)
 	}
 
-	yamlConf, err := config.NewYamlIfExists(opts.configPath)
+	yamlConf, err := config.NewReaderIfExists(opts.configPath)
```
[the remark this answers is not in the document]
sfluor: In fact we either have to call `ddconfig.Load` from https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config.go#L429
Or use the `ReadConfig` method and giving the yaml configuration as an `io.Reader` (which is what we do)

config/config.go
```diff
@@ -221,153 +251,54 @@ func NewDefaultAgentConfig() *AgentConfig {
 		}
 	}
 
-	if isRunningInKubernetes() {
+	if ddconfig.IsKubernetes() {
 		ac.ContainerBlacklist = defaultKubeBlacklist
```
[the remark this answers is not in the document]
sfluor: It looks like ContainerWhitelist is not used either should we remove it too ?

sunhay:
Closing this!

PR #211, opened 2018-11-19
[proc] add network address information to container object

shang-wang:
This PR adds network information to process-agent container object. @DataDog/burrito


proto/agent.proto
```diff
@@ -168,6 +168,13 @@ enum ContainerHealth {
 	unhealthy = 3;
 }
 
+// ContainerNetworkAddress records the IPs, Ports and Protocols for each container
+message ContainerNetworkAddress {
+	string ip = 1;
+	int32 port = 2;
+	string protocol = 3; // tcp/udp
```
sunhay: Should we change this to `ConnectionType` that we have defined below?

shang-wang: Sure. I didn't see it.

checks/container.go
```diff
@@ -152,6 +148,19 @@ func fmtContainers(
 	return chunked
 }
 
+// convertAddressList converts AddressList into process-agent ContainerNetworkAddress objects
+func convertAddressList(ctr *containers.Container) []*model.ContainerNetworkAddress {
+	addrs := make([]*model.ContainerNetworkAddress, 0, len(ctr.AddressList))
+	for _, a := range ctr.AddressList {
+		addrs = append(addrs, &model.ContainerNetworkAddress{
+			Ip:       a.IP.String(),
```
kevinconaway: Will this work with IPv6 addresses as well?

proto/agent.proto
```diff
@@ -168,6 +168,13 @@ enum ContainerHealth {
 	unhealthy = 3;
 }
 
+// ContainerNetworkAddress records the IPs, Ports and Protocols for each container
+message ContainerNetworkAddress {
+	string ip = 1;
```
leeavital: Since this is repeated inside container, when will a container have multiple IPs ?

leeavital: We spoke offline, a container will have a different ip if it is assigned a different network namespace

sunhay:
I think we may have talked about potentially sending this metadata through another endpoint since it will not change that often - is there still plans to do that @shang-wang?

shang-wang:
The container network information is indeed immutable and shouldn't be sent every time. But it's a small list in most cases so it's not so much a burden for now. But in case it causes any problem going forward we will definitely cut it off and send them only when containers are created.

leeavital:
I did a bit of testing. This does indeed add container IPs to container object, but only if:

- the kubelet collector is set (the container IP gathering is not implemented in the docker collector yet)
- the kubelet is set up to allow anonymous auth without SSL. This is the case on GKE, but not on other k8s installs I tested.  Extra configuration is required to successfully connect to the local kubelet.

proto/agent.proto
```diff
@@ -325,10 +333,19 @@ message Connection {
...
+	bytes byteKey = 3;
+	repeated string tags = 4;
+}
+
 message Addr {
-	Host host = 1;
+	Host host = 1;  // post-resolution field
 	string ip = 2;
 	int32 port = 3;
+	ContainerMetadata container = 4; // post-resolution field
```
sunhay: I'm a little worried about this adding this into Addr (as well as the pre-existing `Host`).

Since these objects contain string array tags, there's a high likelihood for duplication, which we've seen cause decoding performance issues for Processes as the message sizes increase quite a bit.

What do you think of changing this `Addr` object to store the `host_id` and/or `container_id`, and then top-level in `CollectorConnections`, we store two maps, `map<host_id, Host>` and `map<container_id, ContainerMetadata>`?

leeavital: I think that makes sense.  Removing the host field from addr is going to break some downstream code (none of which is critical, but it's still some code we'll need to rewrite)

leeavital: Done, I made this field deprecated. It can later become reserved.

proto/agent.proto
```diff
@@ -47,6 +47,14 @@ message CollectorConnections {
 	// Message batching metadata
 	int32 groupId = 5;
 	int32 groupSize = 6;
+
+	// post-resolution field; all the hosts referenced in `connections` keyed by
+	// hostId
+	map<int32, Host> connectedHosts = 7;
```
sunhay: `resolvedHosts`? Can/should we also include the above top-level `Host` in this map?

sunhay: `hostsById`?

proto/agent.proto
```diff
@@ -195,6 +210,7 @@ message Container {
 	int64 started = 24;
 	bytes byteKey = 25;
 	repeated string tags = 26;
+	repeated ContainerAddr addressList = 27;
```
sunhay: nit: `addressList` -> `addresses`

proto/agent.proto
leeavital: I prefer `resolvedHosts`. `hostsById` is sort of redundant since the type explains it. I'll also rename `connectedContainers` 

> Can/should we also include the above top-level Host in this map?

that host **should** be referenced in the connections list,  but we can make it explicit that it should be included.

PR #215, opened 2018-11-21
[proc] put containers and processes running in containers in the same chunk

shang-wang:
To: @DataDog/burrito 

When we split big payloads into smaller chunks, there was a bug that processes for some containers ended up in a different chunk, this results in that when we tried to attach container tags to processes, sometimes the lookup failed so the container tags are missing from those processes.

This PR tries to gather all processes running in containers together with the containers in a single message, while splitting non-container processes into chunks of messages. 

Previously `fmtProcesses` and `fmtContainers` create process/container objects and also splitting them evenly, in this version they only return a list of objects, then pass them along to `chunkProcesses` and `chunkContainers` for splitting.

The change is everywhere and I have next to no confidence that this would work, please review with all your strength.

### DIFF
--- a/checks/container.go
+++ b/checks/container.go
@@ -61,7 +61,7 @@ func (c *ContainerCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Me
 	if len(ctrList) != cfg.MaxPerMessage {
 		groupSize++
 	}
-	chunked := fmtContainers(ctrList, c.lastRates, c.lastRun, groupSize)
+	chunked := chunkContainers(ctrList, c.lastRates, c.lastRun, groupSize)
 	messages := make([]model.MessageBody, 0, groupSize)
 	totalContainers := float64(0)
 	for i := 0; i < groupSize; i++ {
@@ -83,18 +83,8 @@ func (c *ContainerCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Me
 	return messages, nil
 }
 
-// fmtContainers formats and chunks the ctrList into a slice of chunks using a specific
-// number of chunks. len(result) MUST EQUAL chunks.
-func fmtContainers(
-	ctrList []*containers.Container,
-	lastRates map[string]util.ContainerRateMetrics,
-	lastRun time.Time,
-	chunks int,
-) [][]*model.Container {
-	perChunk := (len(ctrList) / chunks) + 1
-	chunked := make([][]*model.Container, chunks)
-	chunk := make([]*model.Container, 0, perChunk)
-	i := 0
+func fmtContainers(ctrList []*containers.Container, lastRates map[string]util.ContainerRateMetrics, lastRun time.Time) []*model.Container {
+	containers := make([]*model.Container, 0, len(ctrList))
 	for _, ctr := range ctrList {
 		lastCtr, ok := lastRates[ctr.ID]
 		if !ok {
@@ -117,7 +107,7 @@ func fmtContainers(
 			tags = []string{}
 		}
 
-		chunk = append(chunk, &model.Container{
+		containers = append(containers, &model.Container{
 			Id:          ctr.ID,
 			Type:        ctr.Type,
 			CpuLimit:    float32(ctr.CPULimit),
@@ -139,7 +129,22 @@ func fmtContainers(
 			Started:     ctr.StartedAt,
 			Tags:        tags,
 		})
+	}
+	return containers
+}
 
+// chunkContainers formats and chunks the ctrList into a slice of chunks using a specific
+// number of chunks. len(result) MUST EQUAL chunks.
+func chunkContainers(ctrList []*containers.Container, lastRates map[string]util.ContainerRateMetrics, lastRun time.Time, chunks int) [][]*model.Container {
+	perChunk := (len(ctrList) / chunks) + 1
+	chunked := make([][]*model.Container, chunks)
+	chunk := make([]*model.Container, 0, perChunk)
+
+	containers := fmtContainers(ctrList, lastRates, lastRun)
+
+	i := 0
+	for _, ctr := range containers {
+		chunk = append(chunk, ctr)
 		if len(chunk) == perChunk {
 			chunked[i] = chunk
 			chunk = make([]*model.Container, 0, perChunk)
--- a/checks/container.go
+++ b/checks/container.go
@@ -58,7 +58,7 @@ func (c *ContainerCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Me
 	}
 
 	groupSize := len(ctrList) / cfg.MaxPerMessage
-	if len(ctrList) != cfg.MaxPerMessage {
+	if cfg.MaxPerMessage*groupSize != cfg.MaxPerMessage {
 		groupSize++
 	}
 	chunked := chunkContainers(ctrList, c.lastRates, c.lastRun, groupSize)
@@ -83,6 +83,7 @@ func (c *ContainerCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Me
 	return messages, nil
 }
 
+// fmtContainers loops through container list and converts them to a list of container objects
 func fmtContainers(ctrList []*containers.Container, lastRates map[string]util.ContainerRateMetrics, lastRun time.Time) []*model.Container {
 	containers := make([]*model.Container, 0, len(ctrList))
 	for _, ctr := range ctrList {
@@ -133,8 +134,8 @@ func fmtContainers(ctrList []*containers.Container, lastRates map[string]util.Co
 	return containers
 }
 
-// chunkContainers formats and chunks the ctrList into a slice of chunks using a specific
-// number of chunks. len(result) MUST EQUAL chunks.
+// chunkContainers formats and chunks the ctrList into a slice of chunks using a specific number of chunks.
+// len(result) MUST EQUAL chunks.
 func chunkContainers(ctrList []*containers.Container, lastRates map[string]util.ContainerRateMetrics, lastRun time.Time, chunks int) [][]*model.Container {
 	perChunk := (len(ctrList) / chunks) + 1
 	chunked := make([][]*model.Container, chunks)
--- a/checks/process.go
+++ b/checks/process.go
@@ -17,6 +17,7 @@ import (
 
 // Process is a singleton ProcessCheck.
 var Process = &ProcessCheck{}
+var emptyCtrID = ""
 
 // ProcessCheck collects full state, including cmdline args and related metadata,
 // for live and running processes. The instance will store some state between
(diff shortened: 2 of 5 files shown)


checks/container.go
```diff
@@ -139,7 +130,22 @@ func fmtContainers(
...
+	chunked := make([][]*model.Container, chunks)
+	chunk := make([]*model.Container, 0, perChunk)
+
+	containers := fmtContainers(ctrList, lastRates, lastRun)
+
+	i := 0
+	for _, ctr := range containers {
+		chunk = append(chunk, ctr)
 		if len(chunk) == perChunk {
 			chunked[i] = chunk
```
leeavital: why not append to chunked instead of keeping track of `i ` ?

checks/process.go
```diff
@@ -76,28 +77,14 @@ func (p *ProcessCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Mess
...
-		messages = append(messages, &model.CollectorProc{
-			HostName:   cfg.HostName,
-			Info:       p.sysInfo,
-			Processes:  chunkedProcs[i],
-			Containers: chunkedContainers[i],
-			GroupId:    groupID,
-			GroupSize:  int32(groupSize),
-		})
+	if len(procsByCtr) == 0 {
+		return []model.MessageBody{}, nil
```
leeavital: why did this change from nil to an empty slice?

checks/process.go
```diff
@@ -106,28 +93,111 @@ func (p *ProcessCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Mess
...
+			GroupId:   groupID,
+		})
+	}
+
+	// pack all containered processes with containers in a message
+	validIndex := 0
+	ctrProcs := make([]*model.Process, 0)
+	for _, ctr := range containers {
+		// if all processes are skipped for this container, don't send the container
+		if _, ok := procsByCtr[ctr.Id]; !ok {
```
leeavital: does this ever happen?

checks/process.go
```diff
@@ -106,28 +93,111 @@ func (p *ProcessCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Mess
...
+			Info:      sysInfo,
+			Processes: c,
+			GroupId:   groupID,
+		})
+	}
+
+	// pack all containered processes with containers in a message
+	validIndex := 0
+	ctrProcs := make([]*model.Process, 0)
+	for _, ctr := range containers {
```
leeavital: Is there a way to move this filtering out of  `createProcCtrMessages` and pass the result in? It's sort of suprising that this function modifies `containers`

checks/process.go
```diff
@@ -106,28 +93,111 @@ func (p *ProcessCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Mess
...
+		containers[validIndex] = ctr
+		validIndex++
+
+	}
+
+	if len(containers[:validIndex]) > 0 {
+		msgs = append(msgs, &model.CollectorProc{
+			HostName:   cfg.HostName,
+			Info:       sysInfo,
+			Processes:  ctrProcs,
```
leeavital: Isn't this going to be every containerized-process on the host? That's going to be quite large won't it? Could we write a loop like:


```go
var procContainerMessages []CollectorProc
nextMessage . = . CollectorProc{}
nextMessageSize = 0
for _, ctr := containers {
   // skip if no procs for ctr
  size :=  size_of_container + size_of_proc * procsByCtr[ctr.Id]
   if size + nextMessageSize > max {

         procContainerMessages = append(procContainerMessages, nextMessage)
         nextMessage = . {}
         nextMessageSize = 0
    } else {
         nextMessage.Containers = append(...)
         nextMesage.Processes = append(...)
    } 
}
```

and get rid of the previous loop altogether?

checks/process_test.go
```diff
@@ -25,6 +28,120 @@ func makeProcess(pid int32, cmdline string) *process.FilledProcess {
...
+	for i, tc := range []struct {
+		cur, last       map[int32]*process.FilledProcess
+		containers      []*containers.Container
+		maxSize         int
+		blacklist       []string
+		expectedChunks  int
+		totalProcs      int
+		totalContainers int
+	}{
+		// this tests no container case
```
leeavital: nit: I've found it helpful to add comments like this to the test data and use t.Run to create named test cases

checks/container.go
shang-wang: I simply copied over the existing logic, but I think `chunked` is already initialized with a list of empty slices: `chunked := make([][]*model.Container, chunks)`, if you use `append` instead of using index it would just append after the empty ones at the end. And to do use pre-allocation is mostly for efficiency reasons.

checks/process.go
shang-wang: Nothing changes, I don't remember why I did that either......I'll change it back

checks/process.go
shang-wang: If your container is running only one process and that process is blacklisted, would that be the case?

checks/process.go
shang-wang: I slightly disagree. the `containers` here is only used to create messages. Changing it in this function doesn't create any side effects. On the other hand, it's changed at the very end and isolates the logic well. But open to discussion just in case I'm wrong.

checks/process.go
shang-wang: I thought about this as well, but I want to simply the logic on the first run because I think the chance that a host runs 100+ containers is small(our chunk size is 100). But since you wrote the logic and it doesn't seem to be very complex, I might give it a shot.

checks/process_test.go
shang-wang: Cool.....any reference of what you mentioned here?

checks/process.go
shang-wang: Per in-person convo with Lee I'll leave the current logic in place.

checks/container.go
leeavital: I think creating `chunked` with `make([][]*model.Container, 0, chunks)` will have the same efficiency advantages and allow you to use append.

checks/process.go
leeavital: `containers` is a parameter to this function and you're modifying it on line 139.  All I'm saying is that if someone tries to re-use `containers` after passing it to `createProcCtrMessages` they're going to be surprised.

checks/process_test.go
leeavital: pointed out some code offline

checks/process_test.go
```diff
@@ -25,6 +28,123 @@ func makeProcess(pid int32, cmdline string) *process.FilledProcess {
...
+		totalContainers int
+	}{
+		{
+			testName:        "no containers",
+			cur:             map[int32]*process.FilledProcess{p[0].Pid: p[0], p[1].Pid: p[1], p[2].Pid: p[2]},
+			last:            map[int32]*process.FilledProcess{p[0].Pid: p[0], p[1].Pid: p[1], p[2].Pid: p[2]},
+			maxSize:         2,
+			containers:      []*containers.Container{},
+			blacklist:       []string{},
+			expectedChunks:  2,
```
sunhay: I think we should rewrite these tests to actually confirm that the processes are grouped with the correct containers, as expected, to increase confidence in this PR, instead of just using hardcoded pre-calculated result values.

It might also be useful to have some generation helper functions like to `generateProcesses(100)` to test more real world scenarios where `maxSize` is 100.

checks/container.go
```diff
@@ -58,10 +58,10 @@ func (c *ContainerCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Me
 	}
 
 	groupSize := len(ctrList) / cfg.MaxPerMessage
-	if len(ctrList) != cfg.MaxPerMessage {
+	if cfg.MaxPerMessage*groupSize != cfg.MaxPerMessage {
```
kevinjqliu: If `groupSize := len(ctrList) / cfg.MaxPerMessage` and `cfg.MaxPerMessage*groupSize`,
then 
`cfg.MaxPerMessage*groupSize = cfg.MaxPerMessage * len(ctrList) / cfg.MaxPerMessage = len(ctrList)`

checks/process_test.go
shang-wang: @sunhay A test is added to address this.

checks/process.go
shang-wang: OK. I'll create a separate list inside the function and only append to it when needed.

shang-wang: @leeavital this is addressed

checks/container.go
```diff
@@ -139,15 +130,26 @@ func fmtContainers(
...
 			Tags:        tags,
 		})
+	}
+	return containers
+}
+
+// chunkContainers formats and chunks the ctrList into a slice of chunks using a specific number of chunks.
+func chunkContainers(ctrList []*containers.Container, lastRates map[string]util.ContainerRateMetrics, lastRun time.Time, chunks, perChunk int) [][]*model.Container {
+	chunked := make([][]*model.Container, 0, chunks)
+	chunk := make([]*model.Container, 0, perChunk)
```
leeavital: It seems redundant to pass both perChunk and chunk because `perChunk * chunks ~= len(ctrList)`. Can we get away with only passing one?

checks/process.go
```diff
@@ -17,6 +17,7 @@ import (
 
 // Process is a singleton ProcessCheck.
 var Process = &ProcessCheck{}
+var emptyCtrID = ""
```
leeavital: nit: const

checks/process.go
```diff
@@ -106,28 +93,106 @@ func (p *ProcessCheck) Run(cfg *config.AgentConfig, groupID int32) ([]model.Mess
...
+			Info:       sysInfo,
+			Processes:  ctrProcs,
+			Containers: ctrs,
+			GroupId:    groupID,
+		})
+	}
+
+	// fill in GroupSize for each CollectorProc and convert them to final messages
+	messages := make([]model.MessageBody, 0, len(msgs))
+	for _, m := range msgs {
```
leeavital: nit: since you're already looping through the list of messages, it might clean things up to move the bookkeeping for counting procs/ctrs here: 

e.g.


```
for _, m := range msgs {
  m.GroupSize = int32(len(msgs))
  messages = append(messages, m)
  totalProcs +=  len(m.Processes)  
  totocalContainers += len(m.Containers)
}
```

checks/container.go
shang-wang: I did this because before calling `chunkContainers` the calculation is already done once: https://github.com/DataDog/datadog-process-agent/pull/215/files#diff-69c4a56eada3efac057ba71ef68a0aa2R60, it's just a matter of passing additional parameter or do the calculation twice.

checks/process.go
sunhay: Why are we doing this? We still want to send the container even if we're not collecting its processes, no?

shang-wang: Would that cause confusion in the UI when people click a container and expecting processes to show up but never happens? I could change the logic as well and add something in UI to make it obvious.

PR #216, opened 2018-11-25
Adding eBPF code for network-tracer

sunhay:
Moving everything from https://github.com/DataDog/tcptracer-bpf to a subpackage within this repo so we don't need to manage multiple repositories.

```
$ tree ebpf
ebpf
├── LICENSE
├── c
│   ├── bpf_helpers.h
│   ├── tracer-ebpf.c
│   ├── tracer-ebpf.h
│   ├── tracer-ebpf.mk
│   └── tracer-ebpf.o
├── common.go
├── common_test.go
├── config.go
├── ebpf_bytes_match_test.go
├── event.go
├── event_common.go
├── event_common_easyjson.go
├── event_test.go
├── offsetguess.go
├── tracer-ebpf.go
├── tracer.go
├── tracer_test.go
├── tracer_unsupported.go
└── types.go

1 directory, 20 files
```

For now, we're keeping the make files needed to build the eBPF code as its only used in the development process - we should change this to use Rake at some point.


cmd/agent/main.go
```diff
@@ -11,7 +11,7 @@ import (
 
 func main() {
 	flag.StringVar(&opts.configPath, "config", "/etc/datadog-agent/datadog.yaml", "Path to datadog.yaml config")
-	flag.StringVar(&opts.netConfigPath, "network-config", "/etc/datadog-agent/network-tracer.yaml", "Path to network-tracer.yaml config")
+	flag.StringVar(&opts.netConfigPath, "network-config", "/etc/datadog-agent/network-ebpf.yaml", "Path to network-tracer.yaml config")
```
leeavital: Is this a breaking change that we need to call out?

sunhay: oops good catch! this should not have changed

PR #217, opened 2018-11-30
Improvements to dev setup

leeavital:
- install a more recent version of Go and dep
- install some additional dependencies (ruby, and mercurial which is required by dep)
- install a version of protoc that agrees with our Rakefile

### DIFF
--- a/tools/dev_setup.sh
+++ b/tools/dev_setup.sh
@@ -19,10 +19,37 @@ EOD
 
 vagrant up
 
+# fix permissions
+cat <<EOD | vagrant ssh
+sudo chown vagrant /home/vagrant/go
+sudo chown vagrant /home/vagrant/go/src/
+sudo chown vagrant /home/vagrant/go/src/github.com
+EOD
+
+# golang setup
+cat <<EOD | vagrant ssh
+mkdir -p /opt
+wget -q https://dl.google.com/go/go1.11.2.linux-amd64.tar.gz
+sudo tar -xzf go1.11.2.linux-amd64.tar.gz -C /opt
+echo 'export GOPATH=/home/vagrant/go' > ~/.bashrc
+echo 'export PATH=/opt/go/bin:/home/vagrant/go/bin:\$PATH' >> ~/.bashrc
+
+sudo wget https://github.com/golang/dep/releases/download/v0.5.0/dep-linux-386 -O /usr/local/bin/dep -q
+sudo chmod +x /usr/local/bin/dep
+EOD
+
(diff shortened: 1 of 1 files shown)


tools/dev_setup.sh
```diff
@@ -19,10 +19,37 @@ EOD
...
+
+# golang setup
+cat <<EOD | vagrant ssh
+mkdir -p /opt
+wget -q https://dl.google.com/go/go1.11.2.linux-amd64.tar.gz
+sudo tar -xzf go1.11.2.linux-amd64.tar.gz -C /opt
+echo 'export GOPATH=/home/vagrant/go' > ~/.bashrc
+echo 'export PATH=/opt/go/bin:/home/vagrant/go/bin:\$PATH' >> ~/.bashrc
+
+sudo wget https://github.com/golang/dep/releases/download/v0.5.0/dep-linux-386 -O /usr/local/bin/dep -q
```
kevinconaway: Is there no way to get the versions we want through the normal package manager?

leeavital: I would have preferred that too, but unfortunately not inside ubuntu. The recommendations are:

- use brew on osx
- use a `curl http://...install.sh | bash` script which is a little scary
- or use one of the precompiled releases
- `go get` which uses the version found on master, which can sometimes subtly break the latest tagged version

https://github.com/golang/dep#installation

For go and protobuf, the image the VM is based on uses apt repositories which have pretty old releases. Docs for both go and protobuf suggest installing from a tarballs, so I opted to do that instead of finding newer repositories to download from.

shang-wang (approved):

kevinconaway (approved):

tools/dev_setup.sh
```diff
@@ -19,10 +19,37 @@ EOD
...
+sudo chown vagrant /home/vagrant/go
+sudo chown vagrant /home/vagrant/go/src/
+sudo chown vagrant /home/vagrant/go/src/github.com
+EOD
+
+# golang setup
+cat <<EOD | vagrant ssh
+mkdir -p /opt
+wget -q https://dl.google.com/go/go1.11.2.linux-amd64.tar.gz
+sudo tar -xzf go1.11.2.linux-amd64.tar.gz -C /opt
```
conorbranagan: This is different from what we build against for staging and prod (https://github.com/DataDog/datadog-process-agent/blob/master/packaging/build_ddagent_release.sh#L27). Maybe we should keep these in sync so we're always testing with the same version in dev as we will release?

PR #222, opened 2018-12-05
WIP: Merge go code to handle TCP and UDP connections in the same way

sfluor:


ebpf/c/tracer-ebpf.c
```diff
@@ -455,6 +431,8 @@ static int increment_tcp_stats(struct sock *sk, struct tracer_status_t *status,
 		}
 
 		struct ipv4_tuple_t t = {};
+		t.metadata = 1;
```
sunhay: Don't hardcode this. Define something in the header, and explicitly set it for udp

ebpf/tracer.go
```diff
@@ -321,3 +225,17 @@ func loadBPFModule() (*bpflib.Module, error) {
 	}
 	return m, nil
 }
+
+func (t *Tracer) timeoutForConnV4(c *ConnTupleV4) int64 {
+	if c.metadata == 1 {
```
sunhay: `t.metadata&1 == 1` and below

ebpf/c/tracer-ebpf.c
```diff
@@ -28,47 +28,23 @@
...
-
-/* This is a key/value store with the keys being an ipv4_tuple_t for send & recv calls
- * and the values being the struct conn_stats_ts_t *.
- */
-struct bpf_map_def SEC("maps/tcp_stats_ipv4") tcp_stats_ipv4 = {
-	.type = BPF_MAP_TYPE_HASH,
-	.key_size = sizeof(struct ipv4_tuple_t),
-	.value_size = sizeof(struct conn_stats_ts_t),
-	.max_entries = 32768,
+	.max_entries = 65536,
```
sunhay: 👍 in the future we'll want to programmatically change this based on system limits

sfluor: :+1:

ebpf/c/tracer-ebpf.c
```diff
@@ -556,6 +537,8 @@ static int increment_udp_stats(struct sock *sk,
 		}
 
 		struct ipv4_tuple_t t = {};
+		t.metadata = CONN_TYPE_UDP;
+
 		if (!read_ipv4_tuple(&t, status, sk)) {
```
leeavital: It could be nice to add a connectionType parameter to `read_ipv{4,6}_tuple()` and set the field in the function. That way we don't need to remember to initialize each tuple with a connection type

ebpf/c/tracer-ebpf.h
```diff
@@ -47,15 +55,18 @@ struct ipv6_tuple_t {
...
+	// Metadata description:
+	// First bit indicates if the connection is TCP (1) or UDP (0)
+	__u32 metadata; // This is that big because it seems that we atleast need a 32-bit aligned struct
 };
 
-#define TRACER_STATE_UNINITIALIZED 0
-#define TRACER_STATE_CHECKING      1
-#define TRACER_STATE_CHECKED       2
-#define TRACER_STATE_READY         3
+static const __u8 TRACER_STATE_UNINITIALIZED = 0;
```
leeavital: why these changes ?

sfluor: This ensures the value of theses constants can't be changed by error in the code

ebpf/tracer.go
```diff
@@ -84,63 +84,43 @@ func NewTracer(config *Config) (*Tracer, error) {
...
 		return nil, err
 	}
 
 	latestTime, ok, err := t.getLatestTimestamp()
 	if err != nil {
 		return nil, err
-	} else if !ok { // if we haven't yet captured any timestamps, there can be no UDP packets
+	}
+
+	if !ok { // if no timestamps have been captured, there can be no packets
```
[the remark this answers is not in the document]
sfluor: Was mainly to be more clear but I can join them again

ebpf/c/tracer-ebpf.c
sfluor: Good idea, will do that 👍 , however there is structs for which we don't use `read_ipv{4,6}_tuple()` so those will still need to initialize the type manually (when we try to map an ipV6 typle to an ipV4 tuple for instance)

PR #223, opened 2018-12-06
Retrieve the TCP retransmit counts for the network tracer

sfluor:
This PR adds a new `retr_count` field to connection stats to count the number of retransmissions.
We also need to move the PID from the key to the value of the map because most of the times the `bpf_get_current_pid_tgid` will return PID: 0 

(see there for an explanation: https://github.com/iovisor/bcc/blob/master/tools/tcpretrans_example.txt#L14-L18)

To reduce the size of the items in the map we could shrink the metadata field to a `u16` and move it to the value side and shrink the `retr_count` to a `u16` too (but it will then be limited to 65536)


ebpf/c/tracer-ebpf.c
```diff
@@ -447,11 +446,14 @@ static int increment_tcp_stats(struct sock *sk, struct tracer_status_t *status,
 			(*val).send_bytes += send_bytes;
 			(*val).recv_bytes += recv_bytes;
 			(*val).timestamp  = ts;
+			(*val).pid = pid >> 32; // Update the pid
```
sunhay: Why do we need to update it? Is this even possible for this to change

sunhay: That's a new connection.

sunhay: The previous connection will have been closed (and removed from the map)

sfluor:
Some additional comments: 

## Cases

### A process tcp connection stops and an other process restarts with the same t = (sport, saddr, dport, daddr) tuple

#### Current situation

The connection for pid 25 stops and will be caught in the `kprobe/tcp_close` and we will delete the map entry for this connection (pid:25, sport:17891, saddr:127.0.0.1, dport: 80, daddr:123.123.123.123).

Then when the new connection comes up it will be caught in `kprobe/tcp_v4_connect` and will be stored in the map without issue.

#### Moving the pid from the map keys to the map values

If we move the pid from the map keys to the map values the same thing would happen since the pid didn't played any role in the scenario.

#### Using a separate map to store the retransmits

The go code would be responsible for linking the informations from this new retransmitsMap and the map that holds the connections, we would also need to flush the retransmits counts from this map when a `kprobe/tcp_close` for a connection occurs.

The downside of this is that we would have a bigger memory footprint because of the additional map.


### A process udp connection stops and an other process restarts with the same t = (sport, saddr, dport, daddr) tuple

#### Current situation

The connection stops but we can't now it because there is no such `udp_close` event. The connection entry will get cleaned when the UDP timeout will occur.

Then when the new connection comes up it will be caught in `kprobe/udp_sendmsg` or `kprobe/udp_recvmsg` and will be stored in the map (with a different key since we include the pid in the map key, except if it has the same PID and the same source address which is very unlikely).

#### Moving the pid from the map keys to the map values

If the pid is no more in the keys, any UDP connection that is seen with the same (sport, saddr, dport, daddr) will be considered as the previous UDP connection (which is not what we want), so this won't work for UDP.

#### Using a separate map to store the retransmits

Since UDP is not concerned about this this will be the same case as the current situation and it will be the go code responsability to link info from the retransmitsMap and the map that holds the connections.

## Memory considerations

If we decide to store the retransmits in a separate map, we would need such a definition:

```c
// 16 bytes
struct retr_key_t {
	__u32 saddr;
	__u32 daddr;
	__u16 sport;
	__u16 dport;
	__u32 netns;
};

// 4 bytes
struct retr_val_t {
	__u32 retransmits;
};
```

So this will size to `20` bytes per entries, if we have 65536 entries (as an example), this would take 1.25 MB of memory (65536 * 20 / 1024 / 1024).

We are currently using:
- 65536 * (ip_v4_tuple_t = 24 + conn_stats_ts_t = 24) / 1024 / 1024 = `3 MB` to hold ipv4 connections
- 65536 * (ip_v6_tuple_t = 44  + conn_stats_ts_t) = 24 / 1024 / 1024 = `4.25 MB` to hold ipv6 connections

So `7.25 MB` in total, if we decide to store the retransmits this adds to `8.5 MB`

If we merge the v4 and v6 maps we could use only `4.25 MB` to support the same number of connections, and this would add to `5.5 MB` with the retransmits counts.

This would also be the case if we would want to monitor [tail loss probe attempts](https://lwn.net/Articles/542642/) using `kprobe/tcp_send_loss_probe`.

sunhay:
Talked in-person, we're going to start with the separate map implementation.

sfluor:
Notes: 
- Removing multiple times from a BPF map do not throw any error and seems safe.
- There is a tradeoff between using the same `conn_tuple_t` for the `tcp_stats` map and not having to rewrite some of the helpers functions but loosing 4 bytes of space per key (`0.25MB` for 65536 entries) and having a different type without the pid field but having to add helper functions to operate on this new type (to read addresses, etc.)

ebpf/c/tracer-ebpf.c
```diff
@@ -24,8 +24,28 @@
         bpf_trace_printk(____fmt, sizeof(____fmt), ##__VA_ARGS__); \
     })
 
-/* This is a key/value store with the keys being an conn_tuple_t for send & recv calls
- * and the values being the struct conn_stats_ts_t *.
+/* Macro to execute the given expression replacing family by the correct family
+ * Needs sk and status to exist before calling
+ */
+#define handle_family(expr)                                                         \
```
sunhay: ✨

PR #225, opened 2018-12-07
Format network-tracer C code using clang-format

sfluor:


ebpf/c/tracer-ebpf.c
```diff
@@ -20,83 +20,83 @@
...
 /* http://stackoverflow.com/questions/1001307/detecting-endianness-programmatically-in-a-c-program */
-__attribute__((always_inline))
-static bool is_big_endian(void) {
-	union {
-		uint32_t i;
-		char c[4];
-	} bint = {0x01020304};
-
-	return bint.c[0] == 1;
+__attribute__((always_inline)) static bool is_big_endian(void)
```
sunhay: This was my blocker when I tried using `clang-format` a while back - are you able to find any configuration that puts the `__attribute__` on its own line?

sfluor: I will look at that 👍

ebpf/c/tracer-ebpf.c
```diff
@@ -105,689 +105,695 @@ static bool is_big_endian(void) {
...
-		default:
-			// not for us
-			return 0;
-	}
-
-	bpf_map_update_elem(&tracer_status, &zero, &new_status, BPF_ANY);
-
-	return 0;
+__attribute__((always_inline)) static int are_offsets_ready_v4(tracer_status_t* status, struct sock* skp, u64 pid)
+{
```
sunhay: Find a config that doesn't put this on a newline please 🙇

leeavital:
Could we make this a CI check, so PRs cannot merge unless they aren't formatted?

sfluor:
We now have a step to check if the files are correctly formatted (also for go files) and the macros are staying on top of the functions declarations @sunhay @leeavital

Rakefile
```diff
@@ -170,6 +179,17 @@ namespace "ebpf" do
     sh "#{sudo} docker build -t #{DOCKER_IMAGE} -f #{DOCKER_FILE} ."
   end
 
+  desc "Format C code using clang-format"
+  task :cfmt do
+    files = `find ebpf/c -name "*.c" -o -name "*.h"`.split("\n")
+    files.each do |file|
+      sh "clang-format -i -style='{BasedOnStyle: WebKit, BreakBeforeBraces: Attach}' #{file}"
+      # This only works with gnu sed
+      sh "sed -i 's/__attribute__((always_inline)) /__attribute__((always_inline))\\
```
leeavital: Is there a clang format rule we can use for this?  The GNU-only sed thing is unfortunate, but I guess acceptable since everyone devving on the C code will need to do so in a VM

sfluor: I did not found a format rule for this :/ a workaround would be to wrap the macro around clang ignore comments.

To make it work on macos it just needs a `brew install gnu-sed --with-default-names`

PR #226, opened 2018-12-10
Skip process info query for pid 0

kevinjqliu:
Windows process agent is very noisy with this specific log line.
image
process agent constantly updates process information. It will query for PID 0 System Idle Process which cannot be queried.
https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-openprocess
`If the specified process is the System Process (0x00000000), the function fails and the last error code is ERROR_INVALID_PARAMETER. `
Which is the same error log we see. 

https://stackoverflow.com/questions/3232401/windows-pid-0-valid

### DIFF
--- a/checks/allprocesses_windows.go
+++ b/checks/allprocesses_windows.go
@@ -356,6 +356,10 @@ func rebuildProcessMapFromWMI() {
 	}
 
 	for pid, proc := range wmimap {
+		if pid == 0 {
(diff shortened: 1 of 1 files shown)


sunhay:
@derekwbrown if you could also take a look for sanity :)

kevinjqliu:
Tested on windows vm, confirmed not logging same error message

PR #227, opened 2018-12-11
Merge the network tracer IPv4 and IPv6 maps

sfluor:
Motivation:

Decrease memory footprint and simplify the code for handling different connections.
Since we would maybe want to add another map to hold connections metrics like tail loss probe or tcp retransmits this allow us to gain some memory to do that.

The issue right now is that if we want to support `N` entries we have to create 2 maps with `N` entries instead of creating only one with `N` entries. 

## Memory considerations

If we decide to store the retransmits in a separate map, we would need such a definition:

We are currently using:
- 65536 * (ip_v4_tuple_t = 24 + conn_stats_ts_t = 24) / 1024 / 1024 = `3 MB` to hold ipv4 connections
- 65536 * (ip_v6_tuple_t = 44  + conn_stats_ts_t) = 24 / 1024 / 1024 = `4.25 MB` to hold ipv6 connections

So `7.25 MB` in total.

By merging the maps (and keeping 65536 entries) we are down to only `4.25MB` so we have a -40% gain in memory usage.


sunhay:
Can you add some details in the PR description, re: cost in space of doing this, etc.

ebpf/c/tracer-ebpf.c
```diff
@@ -411,6 +372,13 @@ static int read_ipv6_tuple(ipv6_tuple_t* tuple, tracer_status_t* status, struct
     tuple->netns = net_ns_inum;
     tuple->metadata = type;
 
+    // Check if we can map IPv6 top IPv4
+    if (is_ipv4_mapped_ipv6(saddr_h, saddr_l, daddr_h, daddr_l) && family == CONN_V6) {
```
sunhay: Flip these conditions - `family == CONN_V6 && is_ipv4_mapped_ipv6(saddr_h, saddr_l, daddr_h, daddr_l)`

ebpf/c/tracer-ebpf.c
```diff
@@ -420,136 +388,47 @@ static int read_ipv6_tuple(ipv6_tuple_t* tuple, tracer_status_t* status, struct
...
-
-    return 0;
-}
-
-__attribute__((always_inline))
-static int increment_udp_stats(struct sock* sk,
+static int increment_conn_stats(struct sock* sk,
     tracer_status_t* status,
     u64 pid_tgid,
+	__u8 type,
```
sunhay: nit: whitespace here is odd

sfluor: Oh right, it's weird that the CI still passed I will have a look

ebpf/event.go
```diff
@@ -121,3 +85,12 @@ func connType(m _Ctype_uint) ConnectionType {
 	}
 	return TCP
 }
+
+func connFamily(m _Ctype_uint) ConnectionFamily {
+	// Second bit of metadata indicates if the connection is IPv6 or IPv4
+	if m&0x10 == 0 {
```
sunhay: Don't hardcode the mask, just reference the c constant directly

ebpf/c/tracer-ebpf.c
```diff
@@ -411,6 +372,13 @@ static int read_ipv6_tuple(ipv6_tuple_t* tuple, tracer_status_t* status, struct
     tuple->netns = net_ns_inum;
     tuple->metadata = type;
 
+    // Check if we can map IPv6 top IPv4
```
leeavital: s/top/to

ebpf/c/tracer-ebpf.c
```diff
@@ -561,13 +440,13 @@ static int increment_udp_stats(struct sock* sk,
             return 0;
         }
 
-        increment_ipv4_stats(sk, status, pid_tgid, CONN_TYPE_UDP, send_bytes, recv_bytes, ts);
+        update_conn_stats(sk, status, pid_tgid, type, CONN_V4, send_bytes, recv_bytes, ts);
```
leeavital: I got tripped up on the name update_conn_stats vs increment_conn_stats is confusing because it's actually update_conn_stats that is doing the incrementing now.  

Maybe `increment_conn_stats` could have a better name like ` handle_message`.

sfluor:
Just updated the PR description with some context about this change

PR #229, opened 2018-12-14
Explicitly enabling process checks should not disable container checks

leeavital:
The yaml configuration parsing for the process-agent makes it impossible to collect both container and process metrics. To reproduce:

- create a file test.yaml:
- build and run process-agent `.rake build && ./process-agent ./config test.yaml`
- the logs include the line

> 2018-12-14 14:29:46 INFO (collector.go:104) - Starting process-agent for host=vagrant, endpoints=[http://foo.bar], **enabled checks=[process rtprocess]**

According to our docs, this configuration should enable both contain and process checks.


sunhay:
The process check includes the container check so this isn’t needed. 

We support either:
- the containers check (enabled by default)
- the process check (enabled manually, includes the container check)

PR #230, opened 2018-12-14
Rename process-checks to reflect that the process check gathers container information

leeavital:
Changes the names of our checks to reflect that the process check also gathers container information. Hopefully this will save other from making the mistake I made in #229


conorbranagan (approved):

config/yaml_config.go
```diff
@@ -143,15 +143,15 @@ func mergeYamlConfig(agentConf *AgentConfig, yc *YamlAgentConfig) (*AgentConfig,
 	}
 	if yc.Process.Intervals.ContainerRealTime != 0 {
```
leeavital: The intervals are still using the old names, I'm looking for a way to support both the old and new without a lot of copy/paste

PR #233, opened 2018-12-17
Support short lived connections using a perf buffer

sfluor:
Add support for short lived connections using a perf buffer, related to: https://github.com/DataDog/datadog-process-agent/pull/232


sfluor:
Some benchmarks:

Running: `go test -tags linux_bpf -bench BenchmarkTCPClose -run=^$ ./ebpf`

```
goos: linux
goarch: amd64
pkg: github.com/DataDog/datadog-process-agent/ebpf
BenchmarkTCPClose/_50_closes-2          2000000000               0.00 ns/op
BenchmarkTCPClose/_100_closes-2         2000000000               0.00 ns/op
BenchmarkTCPClose/_200_closes-2         2000000000               0.01 ns/op
BenchmarkTCPClose/_250_closes-2         2000000000               0.01 ns/op
BenchmarkTCPClose/_300_closes-2         2000000000               0.01 ns/op
---
Total time spent: -653.03157ms
---
BenchmarkTCPClose/eBPF_50_closes-2      2000000000               0.00 ns/op
BenchmarkTCPClose/eBPF_100_closes-2     2000000000               0.01 ns/op
BenchmarkTCPClose/eBPF_200_closes-2     2000000000               0.02 ns/op
BenchmarkTCPClose/eBPF_250_closes-2     2000000000               0.02 ns/op
BenchmarkTCPClose/eBPF_300_closes-2     2000000000               0.03 ns/op
---
Total time spent with eBPF: -1.316955685s
---
```

ebpf/tracer_test.go
```diff
@@ -559,6 +602,46 @@ func BenchmarkTCPSend(b *testing.B) {
...
+			}
+			c.Write(buf)
+		}
+	}
+
+	return func(b *testing.B) {
+		end := make(chan struct{})
+		server := NewTCPServer(echoOnMessage)
+		server.Run(end)
+		for i := 0; i < count; i++ {
```
sunhay: This is not how you're supposed to use the benchmark library. 

You can't choose how many iterations for it to run. 

You need to use `b.N`, and the benchmarking library will run your test as many times as it can in 1 second, and divide the number of iterations over that period to figure out time per iteration.

You can however, change the how long the benchmark will run for (via `-benchtime 30s`)

sfluor: Thanks ! I just updated the benchmarks 👍

ebpf/tracer_test.go
```diff
@@ -559,6 +601,46 @@ func BenchmarkTCPSend(b *testing.B) {
...
+				b.Fatal(err)
+			}
+
+			c.Write([]byte("hello"))
+			err = c.Close()
+			if err != nil {
+				b.Fatalf("can't close connection: %s", err)
+			}
+		}
+		collect()
```
leeavital: I think these benchmarks are not measuring the overhead of collection. We want the  `collect()` call inside the for loop as that's the part that's doing a lot of work.

```
func bench(b testing.B) {
   t := newTracer()
   // wait for t to initialize
   t.Run()
   go func() {
     server := NewTCPServer()
     server.Run(end)
     for { // some stopping condition
        // send message 
     }
   }
   b.Run("test collect", func(b) {
      for i := 0; i < b.N; i++ {
          collect()
       }
    }
}
```

That approach won't work for the perf buffer version because the work isn't being done by collect() in that version but in a callback. And much of the work is done in whatever goroutine kicked off by InitPerfMap. 

For a fair comparison between this and the perfbuffer version, we might want to run the tracer + nettop for a few minutes and sample CPU/memory instead of writing benchmarks.

sfluor: Yeah I added the collect as a third benchmark to take into account the 2 ways of retrieving the collections, I started to go the way of running the tracer and monitoring what happens at the system level, I will update the PR comments when I have results 👍

ebpf/event_common.go
```diff
@@ -87,3 +89,25 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
...
+	for _, c := range conns {
+		key, err := c.ByteKey(buf)
+		if err != nil {
+			// Skip the connection
+			log.Errorf("could not get byte key for connection %v: %s", c, err)
+			continue
+		}
+
+		// If it's the first time we see this connection add it
+		if _, ok := seen[string(key)]; !ok {
```
kevinconaway: Will taking the `string()` of a `[]byte` give you a correct, consistent key?

ebpf/tracer.go
```diff
@@ -136,16 +203,28 @@ func (t *Tracer) getConnections() ([]ConnectionStats, error) {
 	// Remove expired entries
 	t.removeEntries(mp, tcpMp, expired)
 
-	return active, nil
+	// Add the shortlived connections
+	t.Lock()
+	active = append(active, t.closedConns...)
+	t.closedConns = []ConnectionStats{}
```
leeavital: Can we do the clearing to `closedConns` elsewhere?  This makes it so repeated curls to nettracer.sock will yield different results

ebpf/event_common.go
sunhay: Yes :)

There is also an optimization to get rid of an unneeded allocation in these cases: https://github.com/golang/go/issues/3512

ebpf/tracer.go
sfluor: 👍  I updated the PR to mark the connections as collected when the first get occurs and to add a separate goroutine that would check for collected closed connections every 10s and clean them from the Tracer

ebpf/tracer.go
```diff
@@ -81,11 +92,78 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+			case <-ticker.C:
+				log.Infof("Connection stats: %d lost, %d closed", lostCount, closedCount)
+				closedCount = 0
+				lostCount = 0
+			}
+		}
+	}()
+
+	// Start a goroutine to cleanup the closed connections regularly
+	go func() {
```
sunhay: This connection is leaked on `Stop()`

ebpf/tracer.go
```diff
@@ -81,11 +92,78 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+		var closedCount, lostCount uint64
+		ticker := time.NewTicker(5 * time.Minute)
+
+		for {
+			select {
+			case c, ok := <-closedChannel:
+				if !ok {
+					return
+				}
+				closedCount += 1
```
sunhay: a nit, but `closedCount++`

ebpf/tracer.go
```diff
@@ -25,9 +28,17 @@ var (
...
+)
+
 type Tracer struct {
-	m      *bpflib.Module
-	config *Config
+	m       *bpflib.Module
+	perfMap *bpflib.PerfMap
+	config  *Config
+	// Lock for the closedConns
+	sync.Mutex
```
sunhay: We should give this an explicit name if it serves a specific purpose and not for general locking/unlocking of the tracer

ebpf/tracer.go
```diff
@@ -81,11 +93,79 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+				t.closedConnsLock.Unlock()
+
+			case c, ok := <-lostChannel:
+				if !ok {
+					return
+				}
+				lostCount += c
+
+			case <-ticker.C:
+				log.Infof("Connection stats: %d lost, %d closed", lostCount, closedCount)
```
conorbranagan: Will this be too noisy / not useful to the user at the INFO level? Should be it DEBUG instead?

ebpf/tracer.go
```diff
@@ -136,16 +216,43 @@ func (t *Tracer) getConnections() ([]ConnectionStats, error) {
 	// Remove expired entries
 	t.removeEntries(mp, tcpMp, expired)
 
-	return active, nil
+	// Add the closed connections
+	t.closedConnsLock.Lock()
+	// Mark the connections as collected
```
conorbranagan: nit: not a super-useful comment as it's clear from the code

ebpf/event_common.go
```diff
@@ -58,6 +60,9 @@ type ConnectionStats struct {
 	RecvBytes uint64 `json:"recv_bytes"`
 
 	Retransmits uint32 `json:"retransmits"`
+
+	// collected is used internally to know if this connection stats has been collected or not
```
conorbranagan: I think this could use some additional clarity on the "why" of collected instead of just the "what". Why do we need this field, what purpose does it serve. At a glance it's unclear when this is used and `collected` is a pretty generic name.

ebpf/event_common.go
```diff
@@ -87,3 +92,25 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
...
 }
+
+func removeDuplicates(conns []ConnectionStats) []ConnectionStats {
+	connections := make([]ConnectionStats, 0, len(conns))
+	seen := map[string]struct{}{}
+	buf := &bytes.Buffer{}
+	for _, c := range conns {
+		key, err := c.ByteKey(buf)
+		if err != nil {
+			// Skip the connection
```
conorbranagan: nit: not a super-useful comment

ebpf/event_common.go
```diff
@@ -87,3 +92,25 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
 	}
 	return buffer.Bytes(), nil
 }
+
+func removeDuplicates(conns []ConnectionStats) []ConnectionStats {
+	connections := make([]ConnectionStats, 0, len(conns))
+	seen := map[string]struct{}{}
+	buf := &bytes.Buffer{}
```
conorbranagan: we seem to be mixing slice and map initialization styles in this same code base. It'd be nice to just stick with one style (I think `make(...)` seems to be used elsewhere)

ebpf/tracer_test.go
```diff
@@ -251,13 +304,33 @@ func TestTCPClosedConnectionsAreCleanedUp(t *testing.T) {
...
+	conn, ok := findConnection(c.LocalAddr(), c.RemoteAddr(), connections)
+	assert.True(t, ok)
+	assert.Equal(t, clientMessageSize, int(conn.SendBytes))
+	assert.Equal(t, serverMessageSize, int(conn.RecvBytes))
+	assert.Equal(t, 0, int(conn.Retransmits))
+	assert.Equal(t, os.Getpid(), int(conn.Pid))
+	assert.Equal(t, addrPort(server.address), int(conn.DPort))
+
+	// Wait for the cleanup
+	time.Sleep(closedConnCleanupInterval)
```
conorbranagan: This adds 10s to our test time. Can we force a cleanup instead?

sfluor:
Yes, the benchmarks were not really reliable / accurate (there was a lot of variance even when increasing the number of close events / the benchmark time) so I ended up just trying to update the version on a staging node (that had ~50 tcp close events / second) and see the impact.

We had an increase of the memory usage from 20Mb to 30Mb (likely caused by the fact that we would store the closed connections in the tracer), the CPU usage for the tracer did not increased compared to the previous version nor the global CPU usage, the global memory usage did not increase as well.

ebpf/event_common.go
```diff
@@ -87,3 +94,26 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
...
 }
+
+func removeDuplicates(conns []ConnectionStats) []ConnectionStats {
+	connections := make([]ConnectionStats, 0, len(conns))
+	seen := make(map[string]struct{})
+	buf := &bytes.Buffer{}
+
+	for _, c := range conns {
+		key, err := c.ByteKey(buf)
+
```
sunhay: nit: remove this whitespace

ebpf/tracer.go
```diff
@@ -81,11 +93,79 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+		ticker := time.NewTicker(5 * time.Minute)
+
+		for {
+			select {
+			case c, ok := <-closedChannel:
+				if !ok {
+					return
+				}
+				closedCount++
+				stats := decodeRawTCPConn(c)
```
sunhay: Just inline this into the `append()`

ebpf/tracer.go
```diff
@@ -136,16 +216,42 @@ func (t *Tracer) getConnections() ([]ConnectionStats, error) {
...
+	for i := 0; i < len(t.closedConns); i++ {
+		t.closedConns[i].collected = true
+	}
+	active = append(active, t.closedConns...)
+	t.closedConnsLock.Unlock()
+
+	return removeDuplicates(active), nil
+}
+
+func (t *Tracer) cleanupClosedConns() {
```
sunhay: This is confusing to me - why aren't we just setting `t.closedConns = []ConnectionStats{}` above after we've included them all into active? Is it so that multiple calls to `getActiveConnections()` will include this data?

ebpf/tracer.go
```diff
@@ -136,16 +216,42 @@ func (t *Tracer) getConnections() ([]ConnectionStats, error) {
...
+	t.closedConns = newClosedConns
+	t.closedConnsLock.Unlock()
 }
 
 func (t *Tracer) removeEntries(mp, tcpMp *bpflib.Map, entries []*ConnTuple) {
-	for i := range entries {
-		t.m.DeleteElement(mp, unsafe.Pointer(entries[i]))
+	var e *ConnTuple
+	for _, ent := range entries {
+		e = ent
```
sunhay: I don't quite understand the need for this. Isn't `ent` a pointer already?

sunhay: Did you mean for this to be `entries[i]`?

ebpf/tracer.go
sfluor: Yes this is related to: https://github.com/DataDog/datadog-process-agent/pull/233#discussion_r243665986

ebpf/tracer.go
sfluor: woops right !

ebpf/tracer.go
sunhay: Hmm - I'm not sure this makes sense to me; If we're running this on a 10s timer, isn't there a possibility that there's an inconsistency between `getActiveConnections()` calls which happen at the 10s interval in another binary (which may not align with this one)

sunhay: This is actually an issue that affects the `BytesSinceLastCheck` feature I was working on - keeping this state in process-agent causes correctness issues when restarting the agent, and keeping it in the network-tracer causes correctness issues on multiple curls.

ebpf/tracer.go
```diff
@@ -81,31 +88,95 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+		for {
+			select {
+			case c, ok := <-closedChannel:
+				if !ok {
+					return
+				}
+				closedCount++
+
+				t.state.StoreClosedConnection(decodeRawTCPConn(c))
+
```
sunhay: nit: remove this empty line

ebpf/tracer.go
```diff
@@ -81,31 +88,95 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+			case c, ok := <-lostChannel:
+				if !ok {
+					return
+				}
+				lostCount += c
+
```
sunhay: nit: remove this empty line

ebpf/tracer.go
```diff
@@ -81,31 +88,95 @@ func NewTracer(config *Config) (*Tracer, error) {
...
+			case <-ticker.C:
+				log.Debugf("Connection stats: %d lost, %d closed", lostCount, closedCount)
+				closedCount = 0
```
sunhay: nit: `closedCount,lostCount = 0, 0`

PR #234, opened 2019-01-02
Increase connections batch size

sfluor:
(This also includes changes from https://github.com/DataDog/datadog-process-agent/pull/210, will wait for #210  to get merged before merging this one)

We were previously using the same batch size as for processes/containers but network connections are more lightweight so we can send more connections per batch, according to our intake stats a network message is 3 to 4 times smaller than a process message, hence the increase from 100 to 300, there also is a struct comparison for both messages here: https://gist.github.com/sfluor/a7e1dca57891f020ad4c35c8a603e2c2

A process message size is ~ 5Kb
A connections message size is currently ~ 1.4Kb, with the increase it should be 4.2 Kb


config/config_test.go
```diff
@@ -963,13 +963,33 @@ func TestNetYamlConfig(t *testing.T) {
...
 	assert.Equal(true, conf.EnableNetworkTracing)
 	assert.Equal("/tmp/dummy.sock", conf.NetworkTracerSocketPath)
 	assert.Equal("/tmp/net.log", conf.NetworkTracerLogFile)
+	assert.Equal(250, conf.MaxConnsPerMessage)
+}
+
+func TestTooHighMessagesBatches(t *testing.T) {
+	assert := assert.New(t)
+
+	raw := `
```
sunhay: nit: formatting of these lines look a little off :)

PR #237, opened 2019-01-07
Get the latest tag for windows programmatically

leeavital:
This removes a manual step from our windows build process


sunhay:
This starts to break and get messy if we're making test tags. Should we filter results that don't start with an integer, or are not of form `X.Y.Z`?

shang-wang:
@leeavital unfortunately all the gitlab pipelines has been failing because of this somehow. I'll revert this to unblock the release but we can fix this together later.

PR #238, opened 2019-01-10
[network-tracer] Fix C.uint -> uint type casting for go version prior to 1.11 and update Dockerfile-nettop go version to 1.11.2

sfluor:
When trying to build nettop on go 1.9.1:

We get:
```
../../ebpf/event.go:59: cannot use t.metadata (type C.__u32) as type C.uint in argument to connFamily
../../ebpf/event.go:62: cannot use t.metadata (type C.__u32) as type C.uint in argument to connType
../../ebpf/tracer.go:207: cannot use c.metadata (type C.__u32) as type C.uint in argument to connType
```

### DIFF
--- a/ebpf/event.go
+++ b/ebpf/event.go
@@ -56,10 +56,10 @@ func (cs *ConnStatsWithTimestamp) isExpired(latestTime int64, timeout int64) boo
 }
 
 func connStats(t *ConnTuple, s *ConnStatsWithTimestamp, tcpStats *TCPStats) ConnectionStats {
-	family := connFamily(t.metadata)
+	family := connFamily(uint(t.metadata))
 	return ConnectionStats{
 		Pid:         uint32(t.pid),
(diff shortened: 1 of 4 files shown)


ebpf/event.go
```diff
@@ -84,15 +84,15 @@ func ipString(addr_h, addr_l uint64, family ConnectionFamily) string {
...
 	if m&C.CONN_TYPE_TCP == 0 {
 		return UDP
 	}
 	return TCP
 }
 
-func connFamily(m _Ctype_uint) ConnectionFamily {
+func connFamily(m uint) ConnectionFamily {
 	// Second bit of metadata indicates if the connection is IPv6 or IPv4
 	if m&C.CONN_V6 == 0 {
```
sunhay: Does this need to be updated regarding bit-shifting (or is that in another PR)

PR #239, opened 2019-01-10
[network-tracer] Support stateful data for connections

sfluor:
Warning: this will not be merged to master but into https://github.com/DataDog/datadog-process-agent/pull/233
This PR adds a store for connections in the network tracer
This allows to:
- compute relative stats (how much bytes were sent for a given connection since the last time we checked, same with received bytes, retransmits and closed connections)
- retrieve the connections multiple times from multiple "clients" without interfering with other clients store

Edit more context:

## Problem

How to store stateful data in the network tracer for multiple clients:
- Stats since last check to allow easy aggregation on server side
- Closed connections to ensure each client get all the connections that are closed

## Internals

There is a network state that can handle the following operations:

```go
type NetworkState interface {
	Connections(clientID string) []ConnectionStats // Returns the list of connections for the given client
	StoreConnections(conns []ConnectionStats)      // Store new connections in state
	StoreClosedConnection(conn ConnectionStats)    // Store a new closed connection
	RemoveClient(clientID string)                  // Stop tracking stateful data for the given client
}
```

The behavior we want is:
- We want to automatically cleanup the clients if they did not requested anything for a "long" time (5 mn for now)
- We want to keep the closed connections in memory for a client until this client collects it or this client gets removed
- If it's the first time we collect a connection, the stats since last check should be equal to the monotonic value (and not to 0)

### Connections

This returns the connections for a given client, if the client is not registered yet, we register it and return the connections we have stored in the global state.

If the client is already registered we get:
- the connections stored in the global state
- the cached stats since last check for this client

And we merge those two to compute the new stats since last check and update the cached stats since last check for this client

If the client is already registered we also get the list of closed connections it did not collected yet and also update last stats for those, we finally cleanup the closed connections for this client 

### StoreConnections

This stores new connections in the global state and update the overrideConnections for each client if needed (see the Why do we need to store overrideConnections for more detail)

### StoreClosedConnection

This stores a new closed connection in each registered client state and update the overrideConnections for each client if needed (see the Why do we need to store overrideConnections for more detail)

### RemoveClient

This just removes a client from the registered clients (with all its state)

### Why do we need to store sendRecvStats ?

For each client we store a connectionKey => sendRecvStats, to compute the stats since last check

### Why do we need to store overrideConnections ?

We need to store some overrides in the case where a connection closes and a new connection with the exact same key starts before a client collects the closed connection, this allow us to aggregate both connection and make them appear as one connection.


cmd/network-tracer/tracer.go
```diff
@@ -63,8 +64,22 @@ func (nt *NetworkTracer) Run() {
 	http.HandleFunc("/status", func(w http.ResponseWriter, req *http.Request) {})
 
 	http.HandleFunc("/connections", func(w http.ResponseWriter, req *http.Request) {
-		cs, err := nt.tracer.GetActiveConnections()
 
+		// We require to send a client_id to the network state
+		rawCID := req.URL.Query().Get("client_id")
+		clientID := ebpf.DEBUGCLIENT
+
+		if rawCID != "" {
```
sunhay: `if rawCID := req.URL.Query().Get("client_id"); rawCID != ""`

^ since its not being used outside this scope

ebpf/event_common.go
```diff
@@ -95,22 +94,31 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
 	return buffer.Bytes(), nil
 }
 
+// StrKey returns a unique key for this connection represented as a string
+func (c ConnectionStats) StrKey() (string, error) {
+	buf := &bytes.Buffer{}
```
sunhay: There was a reason this was pulled out and reused, it can be pretty expensive to reallocate a buffer every time.

sunhay: https://golang.org/src/bytes/buffer.go

Take a look at: `bootstrap [64]byte // memory to hold first slice; helps small buffers avoid allocation.`. Quite a lot of allocations if you create a new one each iteration.

checks/net.go
```diff
@@ -26,6 +27,7 @@ type ConnectionsCheck struct {
 	// Local network tracer
 	useLocalTracer bool
 	localTracer    *ebpf.Tracer
+	pid            int // Used as a client ID for the local tracer
```
sunhay: Call this `tracerClientId`?

cmd/network-tracer/tracer.go
```diff
@@ -63,8 +64,22 @@ func (nt *NetworkTracer) Run() {
 	http.HandleFunc("/status", func(w http.ResponseWriter, req *http.Request) {})
 
 	http.HandleFunc("/connections", func(w http.ResponseWriter, req *http.Request) {
-		cs, err := nt.tracer.GetActiveConnections()
 
+		// We require to send a client_id to the network state
```
sunhay: Nit: remove the whitespace above this

cmd/network-tracer/tracer.go
```diff
@@ -63,8 +64,22 @@ func (nt *NetworkTracer) Run() {
...
+		rawCID := req.URL.Query().Get("client_id")
+		clientID := ebpf.DEBUGCLIENT
+
+		if rawCID != "" {
+			var err error
+			clientID, err = strconv.Atoi(rawCID)
+			if err != nil {
+				log.Errorf("wrong clientID: '%s': %s", rawCID, err)
```
sunhay: Why not just let `clientId` be a string to completely remove the error potential of this?

ebpf/state.go
```diff
@@ -0,0 +1,209 @@
...
+func (ns *networkState) Connections(id int) []ConnectionStats {
+	ns.connsMutex.Lock()
+	defer ns.connsMutex.Unlock()
+
+	if old := ns.newClient(id); !old {
+		// First time we see this client, use global state
+		return ns.connections
+	}
+
+	conns := removeDuplicates(append(ns.closedConns(id), ns.connections...))
```
sunhay: What happens when a connection is closed + reopened on the same ports during the interval? 

What values do we use + send here? The sum of their counts, the new connections counts, or the old connections counts? What should we send?

ebpf/state.go
```diff
@@ -0,0 +1,209 @@
...
+	ns.connections = conns
+}
+
+// StoreClosedConnection stores the given connection for every client
+func (ns *networkState) StoreClosedConnection(conn ConnectionStats) {
+	ns.clientsMutex.Lock()
+	defer ns.clientsMutex.Unlock()
+
+	for id := range ns.clients {
+		// We only store the pointer to the connection, when it will be cleared for each client it will get GCed
```
sunhay: Good to know 👍

sunhay: Perhaps add this comment above in the client struct as well

ebpf/state.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	return false
+}
+
+func (ns *networkState) RemoveClient(clientID int) error {
+	ns.clientsMutex.Lock()
+	defer ns.clientsMutex.Unlock()
+
+	if _, ok := ns.clients[clientID]; !ok {
+		return fmt.Errorf("can't remove client %d, it is not stored", clientID)
```
sunhay: Thoughts on failing silently here?

ebpf/state.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	delete(ns.clients, clientID)
+	return nil
+}
+
+func (ns *networkState) trackClientExpiry(now time.Time) {
```
sunhay: nit: rename to `removeExpiredClients()`

ebpf/state.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return clients
+}
+
+func (ns *networkState) Connections(id int) []ConnectionStats {
+	ns.connsMutex.Lock()
+	defer ns.connsMutex.Unlock()
+
+	if old := ns.newClient(id); !old {
+		// First time we see this client, use global state
```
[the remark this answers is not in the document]
sfluor: Yes, we start retrieving short lived connections after the first get from the client

sunhay:
Really great work! Some thoughts:

- We should make sure we're handling things correctly when the process-agent first connects to this agent:
  -  Short-lived connections between start + first query are sent
  -  `Last*` counters for connections are returning what transferred over the last interval instead of a default of 0; this one is tricky - feel free to grab me if you'd like to brainstorm on it.

- If a connection closes and re-opens during an interval, perhaps we return the sum of bytes between the two for monotonic + accurate `Last*` counters.

sfluor:
A bit more context:

## Problem

How to store stateful data in the network tracer for multiple clients:
- Stats since last check to allow easy aggregation on server side
- Closed connections to ensure each client get all the connections that are closed

## Internals

There is a network state that can handle the following operations:

```go
type NetworkState interface {
	Connections(clientID string) []ConnectionStats // Returns the list of connections for the given client
	StoreConnections(conns []ConnectionStats)      // Store new connections in state
	StoreClosedConnection(conn ConnectionStats)    // Store a new closed connection
	RemoveClient(clientID string)                  // Stop tracking stateful data for the given client
}
```

The behavior we want is:
- We want to automatically cleanup the clients if they did not requested anything for a "long" time (5 mn for now)
- We want to keep the closed connections in memory for a client until this client collects it or this client gets removed
- If it's the first time we collect a connection, the stats since last check should be equal to the monotonic value (and not to 0)

### Connections

This returns the connections for a given client, if the client is not registered yet, we register it and return the connections we have stored in the global state.

If the client is already registered we get:
- the connections stored in the global state
- the cached stats since last check for this client

And we merge those two to compute the new stats since last check and update the cached stats since last check for this client

If the client is already registered we also get the list of closed connections it did not collected yet and also update last stats for those, we finally cleanup the closed connections for this client 

### StoreConnections

This stores new connections in the global state and update the overrideConnections for each client if needed (see the Why do we need to store overrideConnections for more detail)

### StoreClosedConnection

This stores a new closed connection in each registered client state and update the overrideConnections for each client if needed (see the Why do we need to store overrideConnections for more detail)

### RemoveClient

This just removes a client from the registered clients (with all its state)

### Why do we need to store sendRecvStats ?

For each client we store a connectionKey => sendRecvStats, to compute the stats since last check

### Why do we need to store overrideConnections ?

We need to store some overrides in the case where a connection closes and a new connection with the exact same key starts before a client collects the closed connection, this allow us to aggregate both connection and make them appear as one connection.

sunhay:
> If it's the first time we collect a connection, the stats since last check should be equal to the monotonic value (and not to 0)

On startup of the process-agent, we should probably make a request and throw away the first check results so this doesn't cause issues - can you make the changes to support this?

ebpf/state_test.go
```diff
@@ -0,0 +1,747 @@
...
+		state.StoreClosedConnection(conn2)
+
+		// Second get, we should have monotonic = 8 and last stats = 5
+		conns = state.Connections(client)
+		assert.Equal(t, 1, len(conns))
+		assert.Equal(t, 8, int(conns[0].MonotonicSentBytes))
+		assert.Equal(t, 5, int(conns[0].LastSentBytes))
+	})
+
+	t.Run("TwoShortlivedConnectionsCrossingWithTwoClients", func(t *testing.T) {
```
sunhay: Theses tests look really great 🎉

ebpf/event_common.go
```diff
@@ -95,24 +102,17 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
...
-		}
-
-		// If it's the first time we see this connection add it
-		if _, ok := seen[string(key)]; !ok {
-			connections = append(connections, c)
-			seen[string(key)] = struct{}{}
-		}
-	}
+// ConnByteKeyBuildError represent an error that occurs when building a byte key for a connection
+type ConnByteKeyBuildError struct {
```
sunhay: Can this be deleted? It's not used anywhere

ebpf/state.go
```diff
@@ -0,0 +1,366 @@
...
+	log "github.com/cihub/seelog"
+)
+
+var _ NetworkState = &networkState{}
+
+const (
+	// DEBUGCLIENT is the ClientID for debugging
+	DEBUGCLIENT          = "-1"
+	defaultClientExpiry  = 5 * time.Minute
+	defaultCleanInterval = 10 * time.Second
```
sunhay:
```
defaultClientExpiry  = 2 * time.Minute
defaultCleanInterval = 30 * time.Second
```
^ ?

PR #240, opened 2019-01-14
Add incoming/outgoing connection direction indicator

kevinconaway:
This PR adds the `ConnectionStats.Direction` field which indicates whether the tcp connection is incoming, outgoing or local.  

This is determined by checking if the socket source address and port matches a port/address combination that is in the listening state on the host.

When the agent starts up, we read `/proc/net/tcp` and `/proc/net/tcp6` (these files do not require any special permissions to read) to determine which ports are currently in the listening state.  We also scan the network interfaces to determine the bound IP addresses.

As the tracer runs, we trace the `inet_csk_accept` syscall to track socket accepts.  From that, we grab the listen port and address.  This data is pushed to userspace via a bpf map where the key is the listening port and the value is the state of the socket (initially set to "listening")  When the socket is destroyed, we don't remove it from the bpf map, we instead update a state value to indicate that the port was closed.  This will allow the userspace side to still read the data from the map to determine that the connection _was_ incoming.  The userspace side will delete closed entries from the map.

When we are building the list of connections, we check if the source address is one of the bound IP addresses and if the source ports matches one of the listening ports on the host and if so, we mark the connection as _incoming_.  If the connection is not incoming, we check to see if both the source and dest addresses are locally bound addresses and if so, we mark the direction as _local_.  All other connections are deemed _outgoing_.

cc @DataDog/burrito


ebpf/event_common.go
```diff
@@ -27,12 +27,21 @@ func (c ConnectionType) String() string {
...
+
 const (
 	// AFINET represents v4 connections
 	AFINET ConnectionFamily = 0
 
 	// AFINET6 represents v6 connections
 	AFINET6 ConnectionFamily = 1
+
+	// INCOMING represents connections inbound to the host
+	INCOMING ConnectionDirection = 0
```
sfluor: Could we declare all those constants only in the C code and import them here ?
Like
```suggestion
	INCOMING ConnectionDirection = C.INCOMING
```

ebpf/event_common.go
```diff
@@ -58,11 +67,18 @@ type ConnectionStats struct {
...
 	Retransmits uint32 `json:"retransmits"`
+
+	Direction ConnectionDirection `json:"direction"`
 }
 
 func (c ConnectionStats) String() string {
-	return fmt.Sprintf("[%s] [PID: %d] [%v:%d ⇄ %v:%d] %d bytes sent, %d bytes received, %d retransmits",
-		c.Type, c.Pid, c.Source, c.SPort, c.Dest, c.DPort, c.SendBytes, c.RecvBytes, c.Retransmits)
+	direction := "incoming"
+	if c.Direction == OUTGOING {
```
sfluor: nit: we could implement `String() string` on ConnectionDirection directly

ebpf/port_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"net"
+	"net/url"
+	"strconv"
+	"testing"
+)
+
+func TestReadInitialState(t *testing.T) {
+	l, err := net.Listen("tcp", ":0")
+	require.NoError(t, err)
+	defer func() { _ = l.Close() }()
```
sfluor: `defer l.Close()` is not working here ? Or is to avoid errors about not checking for the returned error ?

kevinconaway: > Or is to avoid errors about not checking for the returned error ?

Yes, this one.

ebpf/proc_net.go
```diff
@@ -0,0 +1,133 @@
...
+
+const tcpListen = 10
+
+// readProcNet reads a /proc/net/ file and returns a map of port->address for all listening addresses
+func readProcNet(path string) (map[uint16]string, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
```
sfluor: should we `defer f.Close()` ?

ebpf/tracer.go
```diff
@@ -209,3 +233,30 @@ func (t *Tracer) timeoutForConn(c *ConnTuple) int64 {
 	}
 	return t.config.UDPConnTimeout.Nanoseconds()
 }
+
```
sfluor: we should add a comment to describe what `populatePortMapping` returns

ebpf/port.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+// AddMapping indicates that something is listening on the provided address and port
+func (pm *PortMapping) AddMapping(port uint16, address string) {
+	pm.ports[port] = address
+}
+
+// RemoveMapping indicates that the provided port is no longer being listened on
+func (pm *PortMapping) RemoveMapping(port uint16) {
```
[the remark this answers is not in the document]
kevinconaway: They should be removed from the mapping when we clean up the bpf map.  Will fix.

ebpf/port.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ReadInitialState reads the /proc filesystem and determines which pids are currently listening on what ports
+func (pm *PortMapping) ReadInitialState() error {
+	start := time.Now()
+
+	if pm.config.CollectTCPConns {
+		ports, err := readProcNet(path.Join(pm.procRoot, "net/tcp"))
+		if err != nil {
+			return fmt.Errorf("error reading tcp state: %s", err)
```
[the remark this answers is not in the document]
kevinconaway: > We might consider a softer failure where we log it but move on, since direction isn't strictly required (is it?) 
No, I don't think so

> and it would get picked up anyway for newer connections w/ bpf events.
Correct.  We would only miss out on identifying tcp connections that were established before the tracer started.

ebpf/tracer.go
```diff
@@ -81,7 +77,13 @@ func NewTracer(config *Config) (*Tracer, error) {
 		return nil, fmt.Errorf("failed to init module: error guessing offsets: %v", err)
 	}
 
-	return &Tracer{m: m, config: config}, nil
+	// TODO: We don't have the proc root configurable yet
```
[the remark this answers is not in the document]
conorbranagan: Chatted a bit w/ @sunhay on slack about this one. I think the config would need to be added into the network-tracer.yaml file. We generate that config in https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config_template.yaml#L734-L744 https://github.com/DataDog/datadog-agent/blob/master/pkg/config/render_config.go#L67-L70. Ideally we'd have this config come from the same render path where it's also written to the `datadog.yaml` file. This probably falls out of the scope of this PR though but should go in before we release.

ebpf/c/tracer-ebpf.c
```diff
@@ -406,15 +432,13 @@ static int read_conn_tuple(conn_tuple_t* tuple, tracer_status_t* status, struct
     // Check if we can map IPv6 to IPv4
     if (family == CONN_V6 && is_ipv4_mapped_ipv6(saddr_h, saddr_l, daddr_h, daddr_l)) {
         tuple->metadata |= CONN_V4;
+
```
sfluor: should we also reset `saddr_h` and `laddr_h` here ? Just to avoid errors in the future

ebpf/port.go
```diff
@@ -10,57 +10,53 @@ import (
 type PortMapping struct {
 	procRoot string
 	config   *Config
-	ports    map[uint16]string
+	ports    map[uint16]interface{}
```
sfluor: you can use `struct{}` instead of interface

ebpf/port.go
```diff
@@ -0,0 +1,68 @@
...
+				log.Errorf("error reading tcp6 state: %s", err)
+			} else {
+				for _, port := range ports {
+					pm.ports[port] = struct{}{}
+				}
+			}
+		}
+	}
+
+	log.Infof("Read initial pid->port mapping in %s", time.Now().Sub(start))
```
sfluor: should we use log.Debug here ?

ebpf/tracer.go
```diff
@@ -26,8 +23,10 @@ var (
 )
 
 type Tracer struct {
-	m      *bpflib.Module
-	config *Config
+	m              *bpflib.Module
+	config         *Config
+	portMapping    *PortMapping
+	localAddresses map[string]interface{}
```
sfluor: this could be:
```suggestion
	localAddresses map[string]struct{}
```

ebpf/port.go
kevinconaway: What do we do elsewhere? Its only logged once when the tracers starts.

proto/agent.proto
```diff
@@ -353,8 +353,13 @@ message Connection {
...
 	uint64 totalBytesSent = 13;
 	uint64 totalBytesReceived = 14;
+	uint32 totalRetransmits = 15;
 
-	ConnectionDirection direction = 15;
+	// Relative counters since last check
+	uint64 lastBytesSent = 16;
+	uint64 lastBytesReceived = 17;
+	uint32 lastRetransmits = 18;
+    ConnectionDirection direction = 19;
```
sfluor: nit: indentation is a bit off

ebpf/c/tracer-ebpf.c
```diff
@@ -118,6 +118,20 @@ struct bpf_map_def SEC("maps/udp_recv_sock") udp_recv_sock = {
...
+/* This maps tracks listening ports. Entries are added to the map via tracing the inet_csk_accept syscall.  The
+ * key in the map is the port and the value is a flag that indicates if the port is listening or not.
+ * When the socket is destroyed (via tcp_v4_destroy_sock), we set the value to be "port closed" to indicate that the
+ * port is no longer being listened on.  We leave the data in place for the userspace side to read and clean up
+ */
+struct bpf_map_def SEC("maps/port_bindings") port_bindings = {
+    .type = BPF_MAP_TYPE_HASH,
+    .key_size = sizeof(__u16),
+    .value_size = sizeof(__u8),
+    .max_entries = 32768,
```
sunhay: Should this be 65k like the rest of our maps?

ebpf/c/tracer-ebpf.c
```diff
@@ -391,17 +405,29 @@ static int read_conn_tuple(conn_tuple_t* tuple, tracer_status_t* status, struct
...
+        if (!saddr_l || !daddr_l) {
+            return 0;
+        }
     } else {
         bpf_probe_read(&saddr_h, sizeof(saddr_h), ((char*)skp) + status->offset_daddr_ipv6 + 2 * sizeof(u64));
         bpf_probe_read(&saddr_l, sizeof(saddr_l), ((char*)skp) + status->offset_daddr_ipv6 + 3 * sizeof(u64));
         bpf_probe_read(&daddr_h, sizeof(daddr_h), ((char*)skp) + status->offset_daddr_ipv6);
         bpf_probe_read(&daddr_l, sizeof(daddr_l), ((char*)skp) + status->offset_daddr_ipv6 + sizeof(u64));
+
+        if (!(saddr_h || saddr_l) || !(daddr_h || daddr_l)) {
```
sunhay: Should this be `(saddr_h && sadddr_l)`? I think this breaks the `is_ipv4_mapped_ipv6` logic, which can have 0 for the high bits.

ebpf/c/tracer-ebpf.c
```diff
@@ -802,6 +828,72 @@ int kprobe__tcp_retransmit_skb(struct pt_regs* ctx) {
...
+        __u8 state = PORT_LISTENING;
+
+        bpf_map_update_elem(&port_bindings, &lport, &state, BPF_ANY);
+    }
+
+    return 0;
+}
+
+SEC("kprobe/tcp_v4_destroy_sock")
+int kprobe__tcp_v4_destroy_sock(struct pt_regs* ctx) {
```
sunhay: Do we need an equivalent for v6?

ebpf/c/tracer-ebpf.c
sunhay: Aside: Is it possible that 0 for source and dest addresses are valid? For local traffic perhaps? https://en.wikipedia.org/wiki/0.0.0.0

ebpf/port.go
```diff
@@ -0,0 +1,68 @@
...
+				log.Errorf("error reading tcp6 state: %s", err)
+			} else {
+				for _, port := range ports {
+					pm.ports[port] = struct{}{}
+				}
+			}
+		}
+	}
+
+	log.Debugf("Read initial pid->port mapping in %s", time.Now().Sub(start))
```
sunhay: If this only happens once, might be worth making this `Info`

ebpf/tracer.go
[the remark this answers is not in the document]
sunhay: We should add the required configuration bits needed to this PR.

Can we do something similar to gopsutil's [`HostProc()`](https://github.com/DataDog/gopsutil/blob/dd/internal/common/common.go#L309-L311) function and make sure this [config logic](https://github.com/DataDog/datadog-process-agent/blob/master/config/config.go#L215-L225) gets called applied for the network-tracer?

ebpf/tracer.go
```diff
@@ -216,7 +244,7 @@ func (t *Tracer) removeEntries(mp, tcpMp *bpflib.Map, entries []*ConnTuple) {
 		// We have to remove the PID to remove the element from the TCP Map since we don't use the pid there
 		entries[i].pid = 0
 		// We can ignore the error for this map since it will not always contain the entry
-		t.m.DeleteElement(tcpMp, unsafe.Pointer(entries[i]))
+		_ = t.m.DeleteElement(tcpMp, unsafe.Pointer(entries[i]))
```
sunhay: Whats will the explicit wildcards on the return value? gofmt?

ebpf/c/tracer-ebpf.c
kevinconaway: No, this is called for both v4 and v6 ([noted here](https://github.com/DataDog/datadog-process-agent/pull/240/files#diff-21d8b26d8e1cc76bc43ff4a5a6f541ddR12))

ebpf/port.go
kevinconaway: I originally had it as INFO but @sfluor suggested making it info per the [agent logging guidelines](https://github.com/DataDog/architecture/blob/0bb11840f08298bb3f87be4fcdb948c21ddd3b5e/rfcs/agent-logs/rfc.md#info--default-level)

ebpf/tracer.go
kevinconaway: Not gofmt, warnings about ignoring returned errors.

ebpf/c/tracer-ebpf.c
kevinconaway: > Should this be (saddr_h && sadddr_l)? I think this breaks the is_ipv4_mapped_ipv6 logic, which can have 0 for the high bits.

If the high bits are zero it will evaluate to `!(1 || 0) -> !1 -> false`.  Both values would have to be zero for the statement to evaluate to `!(0 || 0) -> !0 -> true` which allow the early return

kevinconaway: > Is it possible that 0 for source and dest addresses are valid? For local traffic perhaps?

I haven't noticed any valid traffic excluded by this.  This pattern seems to be repeated in other implementations although we may be inadvertently cargo culting something.

sunhay:
This looks good to me. 

Once https://github.com/DataDog/datadog-process-agent/pull/240#discussion_r250044071 is ready, I think this is good to merge :)

ebpf/c/tracer-ebpf.c
```diff
@@ -379,17 +393,29 @@ static int read_conn_tuple(conn_tuple_t* tuple, tracer_status_t* status, struct
     if (family == CONN_V4) {
         bpf_probe_read(&saddr_l, sizeof(u32), ((char*)skp) + status->offset_saddr);
         bpf_probe_read(&daddr_l, sizeof(u32), ((char*)skp) + status->offset_daddr);
+
+        if (!saddr_l || !daddr_l) {
```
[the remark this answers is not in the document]
kevinconaway: The check was previously at the end of the method, I moved it here so that we could exit early and to simplify the process around checking the high/low bits when we have ipv6->ipv4 mapped addresses.

PR #242, opened 2019-01-17
Freddie/network updates

fred-camara:
This is a pull request to make a few quick changes related to the network tracer & network check. 

1. This PR adds logic to handle the network tracer attempting to remove any stale UNIX sockets.

2. This also adds a default value to `createTimeForPID[conn.Pid]` in order to prevent network connections where the PID creation time is not found from being dropped. We'll associate a value of 0 with network connections missing a value for this parameter.

3. The network collection interval was also increased to 30 seconds, so that the timing matches up more closely with expected use cases in the application.


net/uds.go
```diff
@@ -36,6 +36,19 @@ func NewUDSListener(cfg *config.AgentConfig) (*UDSListener, error) {
 		return nil, fmt.Errorf("can't set the socket at write only: %s", err)
 	}
 
+	fileInfo, err := os.Stat(socketPath)
```
sunhay: Can we add tests around this logic? I think the main agent might have some similar tests that we can base ours on.

checks/net.go
```diff
@@ -242,4 +244,4 @@ func connectionPIDs(conns []ebpf.ConnectionStats) []uint32 {
 		pids = append(pids, pid)
 	}
 	return pids
-}
+}
```
sunhay: nit: you need a new line at the end of this file

checks/net.go
```diff
@@ -141,6 +141,7 @@ func (c *ConnectionsCheck) formatConnections(conns []ebpf.ConnectionStats, lastC
 	// Process create-times required to construct unique process hash keys on the backend
 	createTimeForPID := Process.createTimesforPIDs(connectionPIDs(conns))
 
+
```
sunhay: nit: remove this line

checks/net.go
```diff
@@ -149,8 +150,9 @@ func (c *ConnectionsCheck) formatConnections(conns []ebpf.ConnectionStats, lastC
 			continue
 		}
 
+		// default creation time to ensure network connections from short-lived processes are not dropped 
 		if _, ok := createTimeForPID[conn.Pid]; !ok {
-			continue
+			createTimeForPID[conn.Pid] =  0;
```
sunhay: :thumb:

net/uds.go
```diff
@@ -18,6 +18,7 @@ type UDSListener struct {
 
 // NewUDSListener returns an idle UDSListener
 func NewUDSListener(cfg *config.AgentConfig) (*UDSListener, error) {
+
```
sunhay: nit: remove this empty line

PR #244, opened 2019-01-22
Update connection stats with sync_fetch_and_add

leeavital:
Use __sync_fetch_and_add to change connection stats values as described in https://prototype-kernel.readthedocs.io/en/latest/bpf/ebpf_maps.html#kernel-side-ebpf-program 

This should avoid race conditions where two processors are handling messages for the same connection stat and try to increment the send/received bytes count.


ebpf/c/tracer-ebpf.c
```diff
@@ -451,19 +451,16 @@ static void update_conn_stats(
     t.sport = ntohs(t.sport); // Making ports human-readable
     t.dport = ntohs(t.dport);
 
+    // initialize-if-no-exist the connection stat, and load it
+    conn_stats_ts_t empty = {};
+    bpf_map_update_elem(&conn_stats, &t, &empty, BPF_NOEXIST);
     val = bpf_map_lookup_elem(&conn_stats, &t);
+
     // If already in our map, increment size in-place
     if (val != NULL) {
```
leeavital: technically the val != NULL check is always going to be true, but the BPF verifier will reject this code if we don't check

ebpf/c/tracer-ebpf.c
```diff
@@ -451,19 +451,16 @@ static void update_conn_stats(
     t.sport = ntohs(t.sport); // Making ports human-readable
     t.dport = ntohs(t.dport);
 
+    // initialize-if-no-exist the connection stat, and load it
+    conn_stats_ts_t empty = {};
+    bpf_map_update_elem(&conn_stats, &t, &empty, BPF_NOEXIST);
```
sunhay: Isn't this less efficient for cases we're we are just incrementing (which should be the majority of cases where this is called)? We'll be doing two lookups.

leeavital: Yes, but I don't know of a better way to do this.

sunhay: Whats wrong with the previous method? Were we hitting race conditions?

leeavital: There's a race condition with the following sequence:

- thread 1 handles a tcp_send event (a), gets an empty entry
- thread 2 handles a tcp_send event (b), gets an empty entry
 - thread one decides to pupulate the map for the first time, with sent = a.sent
- thread two decides to populate the map for the first time with sent = b.sent


Really we want the entry to be populated with a.sent + b.sent

sunhay: Good to know - were you able to reproduce that bug in tests?

leeavital: I was not. but nothing in the docs indicate that map access is safe across cores so I think this is worth adding.

PR #245, opened 2019-01-24
[network-tracer] Dynamic BPF map sizes

sfluor:
Related `iovisor/gobpf` change: https://github.com/iovisor/gobpf/pull/166


config/config.go
```diff
@@ -83,13 +83,14 @@ type AgentConfig struct {
...
-	NetworkTracerSocketPath  string
-	NetworkTracerLogFile     string
+	EnableNetworkTracing         bool
+	EnableLocalNetworkTracer     bool // To have the network tracer embedded in the process-agent
+	DisableTCPTracing            bool
+	DisableUDPTracing            bool
+	DisableIPv6Tracing           bool
+	NetworkTracerSocketPath      string
+	NetworkTracerLogFile         string
+	NetworkMaxTrackedConnections uint
```
sunhay: `NetworkMaxTrackedConnections` -> `MaxTrackedConnections`?

ebpf/tracer.go
```diff
@@ -386,3 +385,21 @@ func readLocalAddresses() map[string]struct{} {
...
+			MapMaxEntries: int(c.MaxTrackedConnections),
+		},
+		tcpStatsMap.sectionName(): {
+			MapMaxEntries: int(c.MaxTrackedConnections),
+		},
+		portBindingsMap.sectionName(): {
+			MapMaxEntries: int(c.MaxTrackedConnections),
+		},
+		tcpCloseEventMap.sectionName(): {
+			MapMaxEntries: runtime.NumCPU(),
```
kevinconaway: Are there any oddities around containers that we have to be aware of here?  Where `runtime.NumCPU` would be different than how many actual cpus are used by the perf buffer on the ebpf side?

kevinconaway:
Will you please link the required upstream iovisor/gobpf change(s)?

sfluor:
oops sorry, here it is: https://github.com/iovisor/gobpf/pull/166

ebpf/tracer_test.go
```diff
@@ -522,6 +522,46 @@ func TestUDPDisabled(t *testing.T) {
 	doneChan <- struct{}{}
 }
 
+func TestTooSmallBPFMap(t *testing.T) {
```
sunhay: Nice test :)

ebpf/tracer.go
sfluor: I guess we could set it back to 1024 what do you think @kevinconaway  ?

kevinconaway: The perf array has one element per cpu, right? So the value would have to be sized to the max number of CPU cores.

Alternatively, we could try to figure out what the value is on the ebpf side to see if that is more accurate.

sfluor: Or we could try to use the `procfs` from the host in those cases and manually read the number of cores

sfluor: Yes the perf array has one element per cpu.

sfluor: Unfortunately I don't think we can use non-constants as initializers for BPF programs :/
Trying to use `get_nprocs` from `sys/sysinfo.h` results in:
```
ebpf/c/tracer-ebpf.c:78:20: error: initializer element is not a compile-time constant
    .max_entries = get_nprocs(), // This will get overridden at runtime using the number of CPUs of the current host
                   ^~~~~~~~~~~~
1 error generated
```

kevinconaway: Instead of using `runtime.NumCPU` to initialize the value, perhaps make a cgo call to `get_nprocs`?

sfluor: I think this will be the same issue because we are sandboxed by the host in an environment where we can't see / use all the host cores but I will give it a try 👍

sfluor: Added a change there: https://github.com/DataDog/datadog-process-agent/pull/245/commits/368bccf3c33f6e98cc04256a103ef6495114a741 thanks for the help !

sfluor: Per our discussion we are finally just going to set it back to 1024 since it's the safest and since the map only holds `u16` + `u8` per entry

PR #247, opened 2019-02-03
Setting hostId field in Addr struct to int32

sunhay:
The type of this ID should match what we store in our [`resolvedHosts` map](https://github.com/DataDog/datadog-process-agent/blob/master/proto/agent.proto#L54).

Removing the full `Host` object as it's only used post-resolution and we can easily clean it up when bumping the protobuf version on the backend.

@DataDog/burrito


proto/agent.proto
```diff
@@ -363,12 +363,12 @@ message Connection {
 }
 
 message Addr {
-	// post-resolution field; deprecated in favor of hostId
-	Host host = 1  [deprecated = true];
+	// 1 is deprecated
```
leeavital: thoughts on using `reserved` instead of relying on these comments? 

https://developers.google.com/protocol-buffers/docs/proto#reserved

sunhay: Nice - TIL.

Switched over all of our deprecated comments to this field.

PR #249, opened 2019-02-11
fix check count log

shang-wang:
Currently the collector has only one check counter, if we have multiple checks running, the log doesn't distinguish them from one to the other. This PR creates separate counters for each check, each check would be logged in a separate line.  cc: @DataDog/burrito

### DIFF
--- a/cmd/agent/collector.go
+++ b/cmd/agent/collector.go
@@ -74,8 +74,7 @@ func NewCollector(cfg *config.AgentConfig) (Collector, error) {
 
 func (l *Collector) runCheck(c checks.Check) {
 	var runCounter int32
-	rc, ok := l.runCounters.Load(c.Name())
-	if ok {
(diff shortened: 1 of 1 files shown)


cmd/agent/collector.go
```diff
@@ -71,7 +73,15 @@ func NewCollector(cfg *config.AgentConfig) (Collector, error) {
 }
 
 func (l *Collector) runCheck(c checks.Check) {
-	runCounter := atomic.AddInt32(&l.runCounter, 1)
+	var runCounter int32
```
sunhay:
```go
runCounter := int32(1)
if rc, ok := l.runCounters.Load(c.Name()); ok {
		runCounter = rc.(int32) + 1
}
l.runCounters.Store(c.Name(), runCounter)
```

shang-wang: Nice!

PR #251, opened 2019-02-13
Make sure we check/remove pre-existing sockets before the listen call

sunhay:
Fix the dangling socket issue.

Effectively the same thing we do here: https://github.com/DataDog/datadog-agent/blob/master/pkg/dogstatsd/listeners/uds_common.go#L56-L67


sunhay:
TODO: Still need to make sure tests are updated

PR #253, opened 2019-02-15
[proc] connection payload includes container ID for processes

shang-wang:
In order to deal with ephemeral network connections and resolve local addr, we need to store the relationship between connections and containers, using `pid` as a "foreign key".  This is a change from the agent side to send process -> container relationship. 

@DataDog/burrito


sunhay:
This isn't necessarily just for ephemeral network connection though, right? 

We can use this to short-circuit resolution for local addresses since this information is available to us locally already.

shang-wang:
@sunhay Right that's correct.

checks/net.go
```diff
@@ -260,3 +262,16 @@ func connectionPIDs(conns []ebpf.ConnectionStats) []uint32 {
 	}
 	return pids
 }
+
+func connectionPIDs(conns []*model.Connection) []int32 {
```
sunhay: This duplication is odd. Can we just simplify it and have a single function? Or just reuse the previous function?

checks/process.go
```diff
@@ -168,6 +171,16 @@ func chunkProcesses(procs []*model.Process, size int) [][]*model.Process {
 	return chunks
 }
 
+func cidForPid(ctrList []*containers.Container) map[int32]string {
```
sunhay: nit: `containersbyPID` or something more descriptive

checks/process.go
```diff
@@ -308,6 +316,20 @@ func skipProcess(
 	return false
 }
 
+// ctrForPid uses lastCidForPid and filter down only the pid -> cid that we need
+func (p *ProcessCheck) ctrForPid(pids []int32) map[int32]string {
```
sunhay: This duplication also leaves a bad smell. Why can't we just use the above extracted function?

shang-wang: They are different in nature. Previous function takes containers and compose relationship, this function is filtering to only relevant entries in the result. I'm going to change this function name to be something like `filterContainersByPid`.

checks/net.go
shang-wang: That's kinda hard. One takes `ebpf.ConnectionStats` as input and the other takes `*model.Connection`. I can convert both input to be `[]interface` but every time we use this function we have to convert the input into `[]interface`.

PR #258, opened 2019-02-28
Use BPF monotonic clock time in network state management

sunhay:
This replaces the `time.Time` objects that we've been using, which allows:
  - Smaller `stats{}` objects
  - Consistent timings used when expiring connections and their stats objects.

@DataDog/burrito

### DIFF
--- a/ebpf/state.go
+++ b/ebpf/state.go
@@ -27,7 +27,7 @@ const (
 // - sent and received bytes per connection
 type NetworkState interface {
 	// Connections returns the list of connections for the given client when provided the latest set of active connections
-	Connections(clientID string, latestConns []ConnectionStats) []ConnectionStats
+	Connections(clientID string, latestTime uint64, latestConns []ConnectionStats) []ConnectionStats
 
 	// StoreClosedConnection stores a new closed connection
 	StoreClosedConnection(conn ConnectionStats)
@@ -51,7 +51,7 @@ type stats struct {
 	totalRecv        uint64
 	totalRetransmits uint32
 
-	lastUpdate time.Time
+	lastUpdateEpoch uint64
 }
 
 type client struct {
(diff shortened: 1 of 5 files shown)


ebpf/state.go
```diff
@@ -308,7 +314,7 @@ func (ns *networkState) cleanupState(now time.Time, clearExpiredStats, flushStat
 		if c.lastFetch.Add(ns.expiry).Before(now) {
```
kevinconaway: Should `lastFetch` be using the same time source too ?

sunhay: Not it's fine not to, because this is for fetches by another userspace program.

ebpf/state.go
```diff
@@ -119,10 +122,13 @@ func (ns *networkState) getClients() []string {
 // Connections returns the connections for the given client
 // If the client is not registered yet, we register it and return the connections we have in the global state
 // Otherwise we return both the connections with last stats and the closed connections for this client
-func (ns *networkState) Connections(id string, latestConns []ConnectionStats) []ConnectionStats {
+func (ns *networkState) Connections(id string, latestTime uint64, latestConns []ConnectionStats) []ConnectionStats {
 	ns.Lock()
 	defer ns.Unlock()
 
+	// Update the latest known time
+	ns.latestTimeEpoch = latestTime
```
kevinconaway: Do we want to guard against non-monotonic updates here?  I believe I saw that if there was an error fetching the time it wouldn't be updated.

sunhay: I don't think we need to, since if there's an error fetching the time we don't return any active connections at all (just closed connections)

We don't guard against this when we expire TCP/UDP connections either, so I'd like to keep it consistent until I can dig deeper into the out-of-order + integer underflow issues we've been seeing. I have a few suspicions for whats happening (which is related to these monotonic times)

PR #259, opened 2019-02-28
Expose network namespace in connection messages + enable pprof for network-tracer

sunhay:
Changes:
- Expose network namespace in connection messages.
- Enable pprof for network-tracer (we can access this via the unix socket).
- Expose `latestTimeEpoch` in the `debug/stats` endpoint

@DataDog/burrito


kevinconaway:
> Enable pprof for network-tracer (we can access this via the unix socket).

Since the tracer does run as root, should we gate this (and the debug endpoint) behind a configuration flag for customers who don't want this enabled?

proto/agent.proto
```diff
@@ -366,6 +366,8 @@ message Connection {
 	uint64 lastBytesReceived = 17;
 	uint32 lastRetransmits = 18;
 	ConnectionDirection direction = 19;
+
+	uint32 netNS = 20;
```
shang-wang: Can you add a comment on this because `netNS` is an abbreviation?

sunhay:
I think it's important to leave the debug endpoint in without a config flag, for general customer debugging. The unix socket requires fs group permissions so I don't think we're leaking any kind of information.

For the pprof endpoint, I don't mind throwing that behind a config flag, but It's not immediately obvious how we can do that - call pprof init manually? I'll look into it.

PR #261, opened 2019-03-04
Use common agent config package for parsing process-config

sunhay:
This PR is just part 1 of a series. First step is updating the processes YAML config.

- Use the new common config utility package: `github.com/DataDog/datadog-agent/pkg/config`
- Tests were updated to use resource files for example YAML configuration files.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -20,6 +20,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var originalConfig = config.Datadog
+
+func restoreGlobalConfig() {
+	config.Datadog = originalConfig
+}
+
 func TestBlacklist(t *testing.T) {
 	testBlacklist := []string{
 		"^getty",
@@ -251,11 +257,8 @@ func TestDDAgentConfigWithNewOpts(t *testing.T) {
 }
 
 func TestDDAgentConfigBothVersions(t *testing.T) {
-	origcfg := config.Datadog
 	config.Datadog = config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))
-	defer func() {
-		config.Datadog = origcfg
-	}()
+	defer restoreGlobalConfig()
 
 	assert := assert.New(t)
 	// Check that providing process.* options in the dd-agent conf file works
@@ -285,11 +288,8 @@ func TestDDAgentConfigBothVersions(t *testing.T) {
 }
 
 func TestDDAgentConfigYamlOnly(t *testing.T) {
-	origcfg := config.Datadog
 	config.Datadog = config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))
-	defer func() {
-		config.Datadog = origcfg
-	}()
+	defer restoreGlobalConfig()
 
 	assert := assert.New(t)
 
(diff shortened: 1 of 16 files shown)


config/config_test.go
```diff
@@ -250,6 +251,12 @@ func TestDDAgentConfigWithNewOpts(t *testing.T) {
 }
 
 func TestDDAgentConfigBothVersions(t *testing.T) {
+	origcfg := config.Datadog
+	config.Datadog = config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))
```
sfluor: since `config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))` is often used should we have a helper for it ?

sunhay: It's just tests + a one-liner so I didn't feel too strongly.

sfluor: For the `origcfg` we could just store it as a global variable in the test file and have a `resetGlobalConfig` helper

sunhay: Makes sense - I'll do that

config/config_test.go
```diff
@@ -709,38 +607,28 @@ func TestGetProxySettings(t *testing.T) {
...
-		// Fake the os.Setenv("DD_SITE", tc.site)
-		ddconfig.Datadog.Set("site", tc.site)
-		os.Setenv("DD_PROCESS_AGENT_URL", tc.ddURL)
 
-		agentConfig, err := NewAgentConfig(nil, &YamlAgentConfig{}, nil)
-		assert.NoError(err)
-		assert.Equal(tc.expected, agentConfig.APIEndpoints[0].Endpoint.Hostname())
-	}
+	config.Datadog = config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))
+	agentConfig, err := NewAgentConfig(nil, &NetworkAgentConfig{}, "./testdata/TestEnvSiteConfig.yaml")
```
sfluor: I believe this test was to test the behavior when we had `site` set in env variables too as well as the `process_dd_url` variable in the configuration, could we add this back ?

sunhay: Worth doing for a single test, but seems messy to do it X times. I'll add it back for a single case

sfluor: 👍

config/yaml_config.go
```diff
@@ -263,25 +95,149 @@ func mergeNetworkYamlConfig(agentConf *AgentConfig, networkConf *YamlAgentConfig
...
+	a.setCheckInterval(ns, "container_realtime", "rtcontainer")
+	a.setCheckInterval(ns, "process", "process")
+	a.setCheckInterval(ns, "process_realtime", "rtprocess")
+	a.setCheckInterval(ns, "connections", "connections")
+
+	// A list of regex patterns that will exclude a process if matched.
+	for _, b := range config.Datadog.GetStringSlice(key(ns, "blacklist_patterns")) {
+		r, err := regexp.Compile(b)
+		if err != nil {
+			log.Warnf("Ignoring invalid blacklist pattern: %s", b)
```
sfluor: should we `continue` ?

sunhay: Good point - yes.

config/yaml_config.go
```diff
@@ -263,25 +95,149 @@ func mergeNetworkYamlConfig(agentConf *AgentConfig, networkConf *YamlAgentConfig
...
+	a.Transport = ddutil.CreateHTTPTransport()
+
 	return nil
 }
+
+func (a *AgentConfig) setCheckInterval(ns, check, checkKey string) {
+	if interval := config.Datadog.GetInt(key(ns, "intervals", check)); interval != 0 {
+		log.Infof("Overriding container check interval to %ds", interval)
+		a.CheckIntervals[checkKey] = time.Duration(interval) * time.Second
+	}
```
sfluor: should we log something if the interval is 0 ?

sunhay: No - default would be zero so we'd always see that log message

PR #264, opened 2019-03-07
Move INI config parsing to common agent config util

sunhay:
There's actually more deletions then additions but the `Gopkg.*` changes add noise as well as moving INI configurations to separate test resource files.

This PR is dependent on: https://github.com/DataDog/datadog-agent/pull/3132

Will bump datadog-agent version when thats merged.

@DataDog/burrito


config/config_test.go
```diff
@@ -441,158 +412,124 @@ func TestDDAgentConfigYamlAndNetworkConfig(t *testing.T) {
...
-	}
-	return pf(&http.Request{})
-}
+func TestGetINIProxySettings(t *testing.T) {
+	config.Datadog = config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))
+	defer restoreGlobalConfig()
+
+	http, https := "http", "https"
 
-func TestGetProxySettings(t *testing.T) {
```
sunhay: This whole set of INI proxy tests was modified - namely there is no default proxy port anymore. 

Both the trace and core agents operate without a default proxy port - not sure why we were so behind/different. I'll investigate a bit more and ask the core-agent team if our implementation has been wrong.

Gopkg.toml
```diff
@@ -1,14 +1,14 @@
...
+  revision = "87fcbdb53630c978b8561d2c2272b285123c3d42"
 
 [[constraint]]
   name = "github.com/DataDog/gopsutil"
   revision = "c8f74f1344dd41bc49ec4d3c2377c526aaedfd20"
 
 [[constraint]]
   name = "github.com/gogo/protobuf"
-  revision = "d76fbc1373015ced59b43ac267f28d546b955683"
+  version = "~1.0.0"
```
sunhay: This is required due to a change in `datadog-agent`. We'll have to test this well.

config/yaml_config.go
```diff
@@ -75,19 +77,33 @@ func (a *AgentConfig) loadNetworkYamlConfig(path string) error {
...
 	return nil
 }
 
 // Process-specific configuration
-func (a *AgentConfig) loadProcessYamlConfig(path string) error {
-	config.Datadog.AddConfigPath(path)
-	if strings.HasSuffix(path, ".yaml") { // If they set a config file directly, let's try to honor that
-		config.Datadog.SetConfigFile(path)
-	}
+func (a *AgentConfig) loadProcessConfig(iniPath, yamlPath string) error {
```
leeavital: Should we add a check to make sure iniPath and yamlPath are not both set, since this will behave unexpectedly in case they are. Or even accept one string and infer the type based on the extension?

config/yaml_config.go
```diff
@@ -75,19 +77,33 @@ func (a *AgentConfig) loadNetworkYamlConfig(path string) error {
 	// Pull additional parameters from the global config file.
 	a.LogLevel = config.Datadog.GetString("log_level")
 	a.StatsdPort = config.Datadog.GetInt("dogstatsd_port")
+	if config.Datadog.IsSet("bind_host") {
```
leeavital: Did this move from somewhere, or was it left out before?

config/config.go
```diff
@@ -216,123 +215,20 @@ func NewDefaultAgentConfig() *AgentConfig {
...
-	var section *ini.Section
-	if agentIni != nil {
-		section, _ = agentIni.GetSection("Main")
-	}
-
-	// Pull from the ini Agent config by default.
-	if section != nil {
-		a, err := agentIni.Get("Main", "api_key")
-		if err != nil {
+	if util.PathExists(yamlPath) { // For Agents >= 6 we will have a YAML config file to use.
```
sfluor: Should we log something if both path exists ? (Since it might be breaking for some people)

config/yaml_config.go
sfluor: Since this takes care of loading the ini config aswell should we rename the file ?

config/config_test.go
```diff
@@ -169,48 +142,60 @@ func TestGetHostname(t *testing.T) {
 }
 
 func TestDDAgentMultiAPIKeys(t *testing.T) {
-	// if no endpoint is given but api_keys are there, match the first api_key
-	// with the default endpoint
+	config.Datadog = config.NewConfig("datadog", "DD", strings.NewReplacer(".", "_"))
```
leeavital: do we need to do this given that `NewAgentConfig` will set the global `config.Datadog` ? Same comment on all of these tests

config/yaml_config.go
```diff
@@ -75,19 +77,33 @@ func (a *AgentConfig) loadNetworkYamlConfig(path string) error {
...
+		iniConfig, err := legacy.GetAgentConfig(iniPath)
+		if err != nil {
+			return err
+		}
+		legacy.FromAgentConfig(iniConfig)
+	} else if yamlPath != "" { // Loading modern agent 6 configuration (.yaml)
+		config.Datadog.AddConfigPath(yamlPath)
+		// If they set a config file directly, let's try to honor that
+		if strings.HasSuffix(yamlPath, ".yaml") {
+			config.Datadog.SetConfigFile(yamlPath)
```
leeavital: `yml` is sometimes used as an extension.  I see similar code was elsewhere in the agent config parsing, so I guess we're intentionally only looking for `yaml`

config/yaml_config.go
sunhay: The behavior will change to prefer the YAML if both exist, similar to how the main + trace agents operate. We're going to list this as a breaking change in our release notes.

config/yaml_config.go
sunhay: It was left out before. Noticed it when moving things over

config/config_test.go
sunhay: We really just want it reset after the last test, but is more of a best practice to add it to each test to ensure state is reset.

config/yaml_config.go
sunhay: Yeah, our `yaml` config example is generated with the `.yaml` suffix. I don't think we've run into any cases where a customer has provided `.yml` or another flavor

config/yaml_config.go
sunhay: I was going to merge this all into `config.go` after the environment piece is done to reduce noise :)

config/config.go
sunhay: Good Idea. I can do this and say we're preferring YAML

Gopkg.toml
kevinconaway: Any way to pin this to a hard version?

sunhay: I'd like to keep it in-sync with `datadog-agent`: 
     https://github.com/DataDog/datadog-agent/blob/master/Gopkg.toml#L41

sunhay:
Closing this as we no longer need it (agent 5 is no longer being built anymore for updated versions)

PR #265, opened 2019-03-10
Network tracer should skip local DNS connections by default + fix pprof endpoint panic

sunhay:
Looking at some nodes with higher connection counts, I noticed that a large number of connections were DNS traffic on the local host. I think it's important to capture external DNS traffic, but we can probably get away with disregarding the noisy local traffic.

This PR modifies the network-tracer to do this by default.

**Example node (Capture over 30s):**
- Before: 21014 unique connections
- After: 7720 unique connections

We also fix a bug around disabling pprof on the network-tracer.

@DataDog/burrito

### DIFF
--- a/cmd/network-tracer/tracer.go
+++ b/cmd/network-tracer/tracer.go
@@ -61,9 +61,19 @@ func CreateNetworkTracer(cfg *config.AgentConfig) (*NetworkTracer, error) {
 
 // Run makes available the HTTP endpoint for network collection
 func (nt *NetworkTracer) Run() {
-	http.HandleFunc("/status", func(w http.ResponseWriter, req *http.Request) {})
+	httpMux := http.DefaultServeMux
 
-	http.HandleFunc("/connections", func(w http.ResponseWriter, req *http.Request) {
+	// If profiling is disabled, then we should overwrite handlers for the pprof endpoints
+	// that were registered in init():
+	// https://github.com/golang/go/blob/5bd88b0/src/net/http/pprof/pprof.go#L72-L78
+	// We can only do this by creating a new HTTP Mux that does not have these endpoints handled
+	if !nt.cfg.EnableDebugProfiling {
+		httpMux = http.NewServeMux()
+	}
+
+	httpMux.HandleFunc("/status", func(w http.ResponseWriter, req *http.Request) {})
+
+	httpMux.HandleFunc("/connections", func(w http.ResponseWriter, req *http.Request) {
 		// We require to send a client_id to the network state
 		var clientID string
 		if rawCID := req.URL.Query().Get("client_id"); rawCID != "" {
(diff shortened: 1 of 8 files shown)


cmd/network-tracer/tracer.go
```diff
@@ -61,9 +61,19 @@ func CreateNetworkTracer(cfg *config.AgentConfig) (*NetworkTracer, error) {
...
-	http.HandleFunc("/status", func(w http.ResponseWriter, req *http.Request) {})
+	httpMux := http.DefaultServeMux
 
-	http.HandleFunc("/connections", func(w http.ResponseWriter, req *http.Request) {
+	// If profiling is disabled, then we should overwrite handlers for the pprof endpoints
+	// that were registered in init():
+	// https://github.com/golang/go/blob/5bd88b0/src/net/http/pprof/pprof.go#L72-L78
+	// We can only do this by creating a new HTTP Mux that does not have these endpoints handled
+	if !nt.cfg.EnableDebugProfiling {
+		httpMux = http.NewServeMux()
```
sunhay: We were getting `panic: http: multiple registrations for /debug/pprof/` with the previous method of disabling the pprof endpoints. This method works and has been tested.

ebpf/tracer.go
```diff
@@ -135,7 +138,12 @@ func (t *Tracer) initPerfPolling() (*bpflib.PerfMap, error) {
 					return
 				}
 				atomic.AddUint64(&t.perfReceived, 1)
-				t.state.StoreClosedConnection(decodeRawTCPConn(conn))
+				cs := decodeRawTCPConn(conn)
+				if t.shouldSkipConnection(&cs, t.determineConnectionDirection(&cs)) {
```
kevinconaway: Is it just a convenience to filter it early here since we filter it in `getConnections`?

sunhay: `getConnections` should probably be renamed to `getActiveConnections`. 

It is possible to centralize this logic into a single place, but filtering closed + active conns before they get into the network `state` reduces unneeded storage.

ebpf/tracer.go
```diff
@@ -154,6 +163,13 @@ func (t *Tracer) initPerfPolling() (*bpflib.PerfMap, error) {
 	return pm, nil
 }
 
+// shouldSkipConnection returns whether or not the tracer should ignore a given connection:
+//  • Local DNS (*:53) requests if configured (default: true)
+func (t *Tracer) shouldSkipConnection(conn *ConnectionStats, direction ConnectionDirection) bool {
```
leeavital: Do you want to check for UDP here?

sunhay: DNS calls can be TCP as well

sunhay: https://en.wikipedia.org/wiki/Domain_Name_System#DNS_protocol_transport

PR #266, opened 2019-03-12
[wip] Agent side connection aggregation for common static ports

sunhay:
On a node with many connections, this took us from 7744 connections to 1778.

Downside is that the monotonic counters are currently not correct, and on the backend, any level of aggregation will not provide a correct rollup count.

We may want to keep the sorting logic & create + send through a map of pid -> common static source/dest ports for use on the backend instead.

### DIFF
--- a/ebpf/event_common.go
+++ b/ebpf/event_common.go
@@ -93,11 +93,14 @@ type ConnectionStats struct {
 
 	// Last time the stats for this connection were updated
 	LastUpdateEpoch uint64 `json:"last_update_epoch"`
+
+	// If this connection is the aggregation of multiple connections, this will be > 0
+	RollUpCount uint16 `json:"rollup_count"`
 }
 
 func (c ConnectionStats) String() string {
 	return fmt.Sprintf(
-		"[%s] [PID: %d] [%v:%d ⇄ %v:%d] (%s) %d bytes sent (+%d), %d bytes received (+%d), %d retransmits (+%d)",
+		"[%s, PID:%d] [%v:%d ⇄ %v:%d] (%s) %dB sent (+%d), %dB received (+%d), %d retransmits (+%d), rollup: %d",
 		c.Type,
 		c.Pid,
 		c.Source,
@@ -108,6 +111,7 @@ func (c ConnectionStats) String() string {
 		c.MonotonicSentBytes, c.LastSentBytes,
 		c.MonotonicRecvBytes, c.LastRecvBytes,
 		c.MonotonicRetransmits, c.LastRetransmits,
+		c.RollUpCount,
 	)
 }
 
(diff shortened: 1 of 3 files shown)


ebpf/event_common.go
```diff
@@ -133,3 +137,27 @@ func (c ConnectionStats) ByteKey(buffer *bytes.Buffer) ([]byte, error) {
...
 }
+
+func max(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
+
+func applyFuncOnRange(cs []ConnectionStats, isSame func(a, b *ConnectionStats) bool, f func([]ConnectionStats)) {
```
sunhay: While this looks kind of gnarly - it actually removes a whole lot of duplication + is easily testable :)

ebpf/state.go
```diff
@@ -275,6 +277,102 @@ func (ns *networkState) mergeConnections(id string, active map[string]*Connectio
...
+			for _, c := range pidRange {
+				staticDest[c.DPort]++
+				staticSource[c.SPort]++
+			}
+			commonSource, commonDest := commonPorts(staticSource), commonPorts(staticDest)
+
+			// Go over IP sub-ranges and aggregate
+			applyFuncOnRange(pidRange,
+				func(a, b *ConnectionStats) bool {
+					return a.Source == b.Source && a.Dest == b.Dest
```
leeavital: do you want to rollup if `a.source == b.source` OR `a.dest == b.dest` ?

sunhay: I would say no, because then we'll merge connections from different IP <> IP pairs

sunhay: This condition creates a slice of all connections that share the same source and dest IPs

sunhay: https://goplay.space/#hiCzzkbEJQ0 here's an example of `applyFuncOnRange()` with integers

leeavital: Ah I thought source was a full address, not just IP

leeavital: That said, with this approach we'd be folding

```
a:10 -> b:200
a:11 -> b:200
a:35 -> b:200

a:40 -> b:400
a:41 -> b:400
...
a:64 -> b:400
```

Into one connection instead of two. I could see this being a problem for redis where it's common to have several of the same process each listening on different ports.

leeavital: ...if you have 25 connections open to each instance which I suppose is not terribly likely

sunhay: Sorry I don't quite understand. We still also fold on common static ports, but only over an (IP, IP) range.

In the above case, we'll create two connections:

```
a:* -> b:200
a:* -> b:400
```

But iff the total number of connections to *:200 (and similarly for *:400) on this process are >= 25

PR #270, opened 2019-03-20
Collect container thread count and limit

xvello:
Update the payload definition and the container checks to collect the two new fields `ThreadCount` and `ThreadLimit`.

These are added for the docker and kubelet backends in https://github.com/DataDog/datadog-agent/pull/3171


shang-wang (approved):

proto/agent.proto
```diff
@@ -295,6 +297,8 @@ message ContainerStat {
 	uint32 key = 17;
 	int64 started = 18;
 	bytes byteKey = 19;
+	float threadCount = 20;
```
shang-wang: I wonder why these are float types?

xvello: Indeed, they should be `uint64`, as the source type is, by bad 🤦‍♂️

shang-wang (approved):
Good to go!

PR #272, opened 2019-03-26
Remove INI config parsing/handling

sunhay:
We're not building new agent 5 versions past 5.32.0, so instead of putting effort to move INI config handling via https://github.com/DataDog/datadog-process-agent/pull/264, we can just delete it and simplify our code base.


cmd/agent/main.go
```diff
@@ -12,7 +12,6 @@ import (
 func main() {
 	flag.StringVar(&opts.configPath, "config", "/etc/datadog-agent/datadog.yaml", "Path to datadog.yaml config")
 	flag.StringVar(&opts.netConfigPath, "network-config", "/etc/datadog-agent/network-tracer.yaml", "Path to network-tracer.yaml config")
-	flag.StringVar(&opts.ddConfigPath, "ddconfig", "/etc/dd-agent/datadog.conf", "Path to dd-agent config")
```
sfluor: How do we handle providing a flag that does not exist for people that used both `ddconfig` and `config` ?

cmd/agent/main_windows.go
```diff
@@ -120,7 +118,6 @@ func EnableLoggingToFile() {
 // main is the main application entry point
 func main() {
 	flag.StringVar(&opts.configPath, "config", defaultConfigPath, "Path to datadog.yaml config")
-	flag.StringVar(&opts.ddConfigPath, "ddconfig", defaultOldConfigPath, "Path to dd-agent config")
```
sfluor: same as above

cmd/agent/main.go
sunhay: I don't quite understand the question. 

If you provide a flag that's not handled by the binary, it should just be ignored.

config/yaml_config.go
```diff
@@ -161,7 +161,7 @@ func (a *AgentConfig) loadProcessYamlConfig(path string) error {
 
 	// The maximum number of processes, or containers per message. Note: Only change if the defaults are causing issues.
 	if maxPerMessage := config.Datadog.GetInt(key(ns, "max_per_message")); maxPerMessage > 0 {
-		if maxPerMessage <= maxPerMessage {
+		if maxPerMessage <= maxMessageBatch {
```
sfluor: 😀

cmd/agent/main.go
sfluor: Just tried I don't think that is the case:

```
$ /opt/datadog-agent/bin/process-agent -ac
flag provided but not defined: -ac
```

sunhay: That's pretty wild...

I'll add it back in but ignore it

PR #273, opened 2019-03-26
Environment variables are moved to shared config utility

sunhay:
Environment variables now override values in the shared config package.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -208,6 +208,22 @@ func NewDefaultAgentConfig() *AgentConfig {
 	return ac
 }
 
+func loadConfigIfExists(path string) error {
+	if util.PathExists(path) {
+		config.Datadog.AddConfigPath(path)
+		if strings.HasSuffix(path, ".yaml") { // If they set a config file directly, let's try to honor that
+			config.Datadog.SetConfigFile(path)
+		}
+
+		if err := config.Load(); err != nil {
+			return err
+		}
+	} else {
+		log.Infof("no config exists at %s, ignoring...", path)
+	}
+	return nil
+}
+
 // NewAgentConfig returns an AgentConfig using a configuration file. It can be nil
 // if there is no file available. In this case we'll configure only via environment.
 func NewAgentConfig(yamlPath, netYamlPath string) (*AgentConfig, error) {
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -164,10 +162,6 @@ func NewDefaultAgentConfig() *AgentConfig {
 		StatsdHost: "127.0.0.1",
 		StatsdPort: 8125,
 
-		// Path and environment for the dd-agent embedded python
-		DDAgentPy:    defaultDDAgentPy,
-		DDAgentPyEnv: []string{defaultDDAgentPyEnv},
```
sunhay: Removing these as we don't need them anymore

config/config.go
```diff
@@ -222,20 +216,41 @@ func NewAgentConfig(yamlPath, netYamlPath string) (*AgentConfig, error) {
 
 	// For Agent 6 we will have a YAML config file to use.
 	if util.PathExists(yamlPath) {
```
sfluor: Should we log an error if this is not the case now that we support only yaml config ?

sunhay: Technically you can start the agent without a config. I'll log a warning though 👍

sfluor: oh but then it could be ran by the network-tracer and not exist nevermind

config/config.go
```diff
@@ -289,12 +309,19 @@ func NewNetworkAgentConfig(yamlPath string) (*AgentConfig, error) {
 	}
 
 	if util.PathExists(yamlPath) {
-		if err := cfg.loadNetworkYamlConfig(yamlPath); err != nil {
+		config.Datadog.AddConfigPath(yamlPath)
```
sfluor: should we turn this segment and the one above into a function since it's used twice ?

config/config.go
```diff
@@ -304,138 +331,52 @@ func NewNetworkAgentConfig(yamlPath string) (*AgentConfig, error) {
...
+		"DD_PROXY_HTTPS":    "proxy.https",
+
+		"DD_LOGS_STDOUT":    "log_to_console",
+		"LOG_TO_CONSOLE":    "log_to_console",
+		"DD_LOG_TO_CONSOLE": "log_to_console",
+		"LOG_LEVEL":         "log_level", // Support LOG_LEVEL and DD_LOG_LEVEL but prefer DD_LOG_LEVEL
+		"DD_LOG_LEVEL":      "log_level",
+	} {
+		if v, ok := os.LookupEnv(envKey); ok {
+			config.Datadog.Set(cfgKey, v)
```
sfluor: I thought the goal was to make `Set` private did it changed ?

sunhay: Yeah, I thought so too. I'm not sure how we'd properly handle environment variables then.

I'm just going to follow the trace-agent's example for now and we can change in the future depending on what happens downstream.

PR #280, opened 2019-03-29
Accept different return path from `GetProgramDataDir()`.  Underlying

derekwbrown:
implementation was changed to be the root of the config path
(programdata\datadog) by default, to allow for changing the config path
at install time.  Amend the return value accordingly, and update the
commit hash from `datadog-agent`

### DIFF
--- a/cmd/agent/main_windows.go
+++ b/cmd/agent/main_windows.go
@@ -38,9 +38,9 @@ func init() {
 	fmt.Printf("main_windows.init()")
 	pd, err := winutil.GetProgramDataDir()
 	if err == nil {
-		defaultConfigPath = filepath.Join(pd, "Datadog", "datadog.yaml")
-		defaultConfdPath = filepath.Join(pd, "Datadog", "conf.d")
-		defaultLogFilePath = filepath.Join(pd, "Datadog", "logs", "process.log")
+		defaultConfigPath = filepath.Join(pd, "datadog.yaml")
+		defaultConfdPath = filepath.Join(pd, "conf.d")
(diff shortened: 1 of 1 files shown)


cmd/agent/main_windows.go
```diff
@@ -39,9 +39,10 @@ func init() {
 	fmt.Printf("main_windows.init()")
 	pd, err := winutil.GetProgramDataDir()
 	if err == nil {
-		defaultConfigPath = filepath.Join(pd, "Datadog", "datadog.yaml")
-		defaultConfdPath = filepath.Join(pd, "Datadog", "conf.d")
-		defaultLogFilePath = filepath.Join(pd, "Datadog", "logs", "process.log")
+		defaultConfigPath = filepath.Join(pd, "datadog.yaml")
+		defaultOldConfigPath = filepath.Join(pd, "datadog.conf")
```
shang-wang: I think you don't have this `defaultOldConfigPath` initialized properly. All other defaultXXX variables have initializations several lines above, so should be `defaultOldConfigPath`.

shang-wang:
I left a comment, should be an easy fix. Also you might have to run `dep ensure` to update `Gopkg.lock` for your change in `Gopkg.toml`.

cmd/agent/main_windows.go
derekwbrown: it looks like at some point the `defaultOldConfigPath` was removed completely?

sunhay:
I'm fixing the merge conflict (I updated to a later version in another PR) and will then merge this PR

PR #281, opened 2019-03-29
[process-agent] Add testing around secret configs

sfluor:


Rakefile
```diff
@@ -55,9 +55,11 @@ end
 desc "Test Datadog Process agent"
 task :test do
   cmd = "go list ./... | grep -v vendor | xargs go test"
+  tags = ["secrets"]
```
leeavital: this build tag does not seem to exist anywhere

sunhay: This should only be done for unix (secrets isn't supported for Windows as it runs as root)

config/testdata/TestDDAgentConfigYamlEnc2.yaml
```diff
@@ -0,0 +1,6 @@
+api_key: "ENC[dangerous_key]"
```
sunhay: nit but maybe `encrypted_key`?

config/testdata/secret.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type Input struct {
+	Version string   `json:"version"`
+	Secrets []string `json:"secrets"`
+}
+
+type SecretOutput struct {
+	Value string `json:"value"`
+	// Use pointer for error so if it's not provided null will be used
```
leeavital: I believe you can add `omitempty` (or something similar) to the json tag to achieve this without using a pointer

config/config_test.go
```diff
@@ -166,6 +168,65 @@ func TestDefaultConfig(t *testing.T) {
 	os.Unsetenv("DOCKER_DD_AGENT")
 }
 
+func setupSecretScript(script string) error {
```
leeavital: since we always use the same executable, consider removing the argument and wrapping this in sync.Once() so we only have to compile the program once

Rakefile
sfluor: I thought it was required but it seems to work without it @sunhay any thought ?

config/testdata/secret.go
sfluor: I think the behavior is different `omitempty` will return a json like `{"value": "abc"}` when we want `{"value": "abc", "error": null}`

Rakefile
sunhay: https://github.com/DataDog/datadog-agent/blob/e98a1266a754ef3fa33973cf7189c7ef1ecb1a46/pkg/secrets/fetch_secret.go

sfluor: Nevermind I had an outdated local version 👍

PR #289, opened 2019-04-09
[network-tracer] Allow to exclude some linux versions

sfluor:
This allows to exclude some linux versions to avoid seeing issues (like `oopses` here: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1763454)


config/config.go
```diff
@@ -70,17 +70,18 @@ type AgentConfig struct {
...
+	EnableDebugProfiling               bool
+	DisableTCPTracing                  bool
+	DisableUDPTracing                  bool
+	DisableIPv6Tracing                 bool
+	CollectLocalDNS                    bool
+	NetworkTracerSocketPath            string
+	NetworkTracerLogFile               string
+	MaxTrackedConnections              uint
+	NetworkBPFDebug                    bool
+	NetworkTracerExcludedLinuxVersions []string
```
sunhay: `NetworkTracerExcludedLinuxVersions` -> `ExcludedBPFLinuxVersions`

ebpf/common.go
```diff
@@ -3,16 +3,95 @@ package ebpf
...
+
+// KERNEL_VERSION(a,b,c) = (a << 16) + (b << 8) + (c)
+// Per https://github.com/torvalds/linux/blob/master/Makefile#L1187
+func linuxKernelVersionCode(major, minor, patch uint32) uint32 {
+	return (major << 16) + (minor << 8) + patch
+}
+
+var ubuntu44119Error = "got ubuntu kernel %s with known bug, see: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1763454"
+var defaultExclusionList = map[uint32]string{
+	linuxKernelVersionCode(4, 4, 119): ubuntu44119Error,
```
sunhay: This doesn't actually confirm that it's ubuntu though. Just the linux kernel itself. Isn't that an issue?

config/yaml_config.go
```diff
@@ -45,6 +45,7 @@ func (a *AgentConfig) loadNetworkYamlConfig(path string) error {
 	}
 
 	a.NetworkBPFDebug = config.Datadog.GetBool(key(netNS, "bpf_debug"))
+	a.ExcludedBPFLinuxVersions = config.Datadog.GetStringSlice(key(netNS, "excluded_linux_versions"))
```
sunhay: You'll want to wrap this with a `isSet` check

ebpf/common.go
sfluor: Indeed the issue seems to be only on Ubuntu, I will add something to check the OS name

config/yaml_config.go
sfluor: Won't it just default to false ?

sunhay: It'll also throw an trace error since there's no default.

sfluor: 👍

ebpf/common.go
sunhay: Might be worth looking at `bpflib.CurrentKernelVersion()`. They do some semi-complicated parsing of the `uname` output

util/util.go
```diff
@@ -93,3 +96,22 @@ func GetDockerSocketPath() (string, error) {
 	}
 	return sockPath, nil
 }
+
+// GetPlatform returns the current platform we are running on by calling "python -mplatform"
+func GetPlatform() (string, error) {
+	cmd := exec.Command("python", "-m", "platform")
```
sunhay: Are you guaranteed to actually have python installed?

sfluor: I think python comes installed in most of the unix distributions, we can still add something to use `lsb_release` if the python command fails

cmd/agent/main_common.go
```diff
@@ -73,10 +73,18 @@ Exiting.`
 )
 
 func runAgent(exit chan bool) {
+	platform, err := util.GetPlatform()
+	if err != nil {
+		log.Errorf("error retrieving platform: %s", err)
```
sunhay: Debug logging for this is better.

ebpf/common.go
```diff
@@ -13,6 +27,72 @@ var (
...
+	currentKernelCode, err := CurrentKernelVersion()
+	if err != nil {
+		return false, fmt.Errorf("could not get kernel version: %s", err)
+	}
+
+	platform, _ := util.GetPlatform()
+	return verifyOSVersion(currentKernelCode, platform, exclusionList)
+}
+
+func verifyOSVersion(currentKernelCode uint32, platform string, exclusionList []string) (bool, error) {
```
sunhay: nit: `currentKernelCode` -> `kernelCode`

sunhay:
Do we need to update the seccomp profile for this?

sfluor:
The seccomp profile is now just the base seccomp profile + the syscalls needed for BPF specific parts so I think it should be fine.

packaging/Dockerfile-ebpf
```diff
@@ -6,7 +6,7 @@ ENV GOPATH /go
 # vim-common is needed for xxd
 # vim-minimal needs to be updated first to avoid an RPM conflict on man1/vim.1.gz
 RUN dnf update -y vim-minimal && \
-	dnf install -y -b llvm clang kernel-devel make binutils vim-common golang go-bindata ShellCheck git file rpm sudo
+	dnf install -y -b llvm clang kernel-devel-4.16.3-301.fc28.x86_64 make binutils vim-common golang go-bindata ShellCheck git file rpm sudo
```
sunhay: Good catch.

Issue #292, opened 2019-04-12
Hardcoded ports 6062 and 6060 - they need to be configurable / documentation updated to reflect ports in use


0pubnub:
Agent ports (profile-server) are hardcoded and not mentioned anywhere in the documentation. This is causing conflicts with running processes listening to these ports. 

- Port 6062:

https://github.com/DataDog/datadog-process-agent/blob/0f8a0eca9719dd5b19d72be844f36757c88a9bfb/cmd/agent/main_common.go#L155-L167

- Port 6060:

https://github.com/DataDog/datadog-log-agent/blob/e3fb0e64f1b5eb11887fd0953606884fadea9053/pkg/logagent/main.go#L54

leeavital:
Hi @0pubnub , thank you for the bug report. 

The two conflicts have different solutions.

For the process agent port, I've opened a pull request to allow the the process agent debug port to be configured with `process_config.expvar_port`. This will land in the next release of the agent.

For the logs agent (port 6060) -- you are running an older 5.x version of the agent. As of 6.0.0, the logs agent runs in process with the core agent and thus uses the same expvar_port as configured in:

https://github.com/DataDog/datadog-agent/blob/master/pkg/config/config_template.yaml#L131-L132

sunhay:
6.11.2 includes fixes for this issue so I'm closing.

PR #293, opened 2019-04-15
Make process_config.expvar configurable

leeavital:
Adds a flag, similar to that of the core agent: `procss_agent.expvar_port` which describes where the default HTTP handler will bind.  

Fixes the process-agent side of #292


sfluor (approved):
Should we add a test case for this ?

PR #297, opened 2019-04-25
Additional endpoints for debugging eBPF maps and general network state

sunhay:
[see: title]


cmd/network-tracer/tracer.go
```diff
@@ -74,53 +74,80 @@ func (nt *NetworkTracer) Run() {
...
 	})
 
 	http.Serve(nt.conn.GetListener(), httpMux)
 }
 
+func getClientID(req *http.Request) string {
+	var clientID = ebpf.DEBUGCLIENT
+	if rawCID := req.URL.Query().Get("client_id"); rawCID != "" {
+		clientID = rawCID
+	}
```
sfluor: nit: thoughts on logging a message when we default to the debug client ?

sunhay: I don't think we'd want the logs since they'll show up every time we do a debug call

PR #298, opened 2019-04-25
Use a re-usable buffer for appending active connections

kevinconaway:
Profiling the tracer shows `getConnections` as the largest source of allocations, specifically building up the active connection list.  For hosts with a high number of connections, we continuously allocate (and grow) a large array every time.

This PR adds a reusable buffer in the tracer for appending the active connections to which coarsely grows or shrinks over time.

I also 8-byte aligned the `ConnectionStats` struct to save a whopping 8 bytes of memory

@DataDog/burrito

### DIFF
--- a/ebpf/tracer.go
+++ b/ebpf/tracer.go
@@ -162,7 +162,7 @@ func (t *Tracer) GetActiveConnections(clientID string) (*Connections, error) {
 	t.bufferLock.Lock()
 	defer t.bufferLock.Unlock()
 
-	latestConns, latestTime, err := t.appendActiveConnections(t.buffer[:0])
+	latestConns, latestTime, err := t.getConnections(t.buffer[:0])
 	if err != nil {
 		return nil, fmt.Errorf("error retrieving connections: %s", err)
 	}
@@ -177,7 +177,7 @@ func (t *Tracer) GetActiveConnections(clientID string) (*Connections, error) {
 	return &Connections{Conns: t.state.Connections(clientID, latestTime, latestConns)}, nil
 }
 
-func (t *Tracer) appendActiveConnections(active []ConnectionStats) ([]ConnectionStats, uint64, error) {
+func (t *Tracer) getConnections(active []ConnectionStats) ([]ConnectionStats, uint64, error) {
 	mp, err := t.getMap(connMap)
 	if err != nil {
 		return nil, 0, fmt.Errorf("error retrieving the bpf %s map: %s", connMap, err)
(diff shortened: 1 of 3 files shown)


ebpf/tracer.go
```diff
@@ -248,7 +262,13 @@ func (t *Tracer) getTCPStats(mp *bpflib.Map, tuple *ConnTuple) *TCPStats {
 	tuple.pid = 0
 
 	stats := &TCPStats{retransmits: 0}
-	_ = t.m.LookupElement(mp, unsafe.Pointer(tuple), unsafe.Pointer(stats))
+
+	// Don't bother looking in the map if the connection is UDP, there will never be data for that and we will avoid
+	// the overhead of the syscall and creating the resultant error
+	if connType(uint(tuple.metadata)) == TCP {
```
sunhay: Nice catch here

ebpf/tracer.go
```diff
@@ -154,15 +159,25 @@ func (t *Tracer) Stop() {
...
+		t.buffer = make([]ConnectionStats, 0, cap(t.buffer)*2)
+	} else if len(latestConns) <= cap(t.buffer)/2 {
+		t.buffer = make([]ConnectionStats, 0, cap(t.buffer)/2)
+	}
+
 	return &Connections{Conns: t.state.Connections(clientID, latestTime, latestConns)}, nil
 }
 
-func (t *Tracer) getConnections() ([]ConnectionStats, uint64, error) {
+func (t *Tracer) appendActiveConnections(active []ConnectionStats) ([]ConnectionStats, uint64, error) {
```
sunhay: The name of this function is a little odd. just `getConnections`?

ebpf/tracer.go
```diff
@@ -154,15 +159,25 @@ func (t *Tracer) Stop() {
 }
 
 func (t *Tracer) GetActiveConnections(clientID string) (*Connections, error) {
-	latestConns, latestTime, err := t.getConnections()
+	t.bufferLock.Lock()
+	defer t.bufferLock.Unlock()
+
+	latestConns, latestTime, err := t.appendActiveConnections(t.buffer[:0])
```
sunhay: Can we maintain buffer locking + growing/shrinking the buffer in `appendActiveConnections`? There's no reason it needs to be exposed

kevinconaway: I believe that the buffer needs to be locked for the duration of the call to `t.state.Connections` since the underlying buffer is potentially being used there.

Otherwise there could be a race condition if `GetActiveConnections` is called concurrently

ebpf/tracer.go
```diff
@@ -154,15 +159,25 @@ func (t *Tracer) Stop() {
...
+		t.buffer = make([]ConnectionStats, 0, cap(t.buffer)*2)
+	} else if len(latestConns) <= cap(t.buffer)/2 {
+		t.buffer = make([]ConnectionStats, 0, cap(t.buffer)/2)
+	}
+
 	return &Connections{Conns: t.state.Connections(clientID, latestTime, latestConns)}, nil
 }
 
-func (t *Tracer) getConnections() ([]ConnectionStats, uint64, error) {
+func (t *Tracer) getConnections(active []ConnectionStats) ([]ConnectionStats, uint64, error) {
```
sfluor: nit: adding a comment to explain why `getConnections` takes a slice of ConnectionStats as argument could help reading this function

PR #303, opened 2019-05-07
[network-tracer] Remove the expiry logic from the state and expose a RemoveConnections

sfluor:
We had underflows occuring because (mostly UDP) connections were cleaned from the eBPF map but not from the userspace state, this should sync the eBPF map and the userspace state.


ebpf/state.go
```diff
@@ -320,49 +304,23 @@ func (ns *networkState) RemoveClient(clientID string) {
...
 
-	// Remove expired clients + stats
-	deletedStats := 0
+	now := time.Now()
+
+	// Remove connections
 	for id, c := range ns.clients {
-		if c.lastFetch.Add(ns.expiry).Before(now) {
+		// Check if we should expire client
+		if c.lastFetch.Add(ns.clientExpiry).Before(now) {
```
sunhay: If no connections are ever expired, then this will never get called... right? Not sure we want to do that

ebpf/tracer.go
```diff
@@ -249,6 +249,13 @@ func (t *Tracer) getConnections(active []ConnectionStats) ([]ConnectionStats, ui
 }
 
 func (t *Tracer) removeEntries(mp, tcpMp *bpflib.Map, entries []*ConnTuple) {
+	var buf bytes.Buffer
```
sunhay: Create this buffer only once in the `Tracer`?

ebpf/state.go
sfluor: They were still getting expired even if `keys` was empty but I will make a separate function 👍

PR #305, opened 2019-05-08
[network-tracer] use human readable keys in dumps

sfluor:
Now:
```
p:345|src:127.0.0.1:4444|dst:192.168.0.103:8888|f:1|t:0
```

Before:
```
"\Y127.0.0.1||192.168.0.103
```

### DIFF
--- a/ebpf/event_test.go
+++ b/ebpf/event_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ebpf/event_test.go
```diff
@@ -23,6 +23,27 @@ var (
...
+			Type:   0,
+			Family: 1,
+			Source: "127.0.0.1",
+			Dest:   "192.168.0.103",
+			SPort:  4444,
+			DPort:  8888,
+		},
+	} {
+		bk, err := c.ByteKey(buf)
+		assert.NoError(t, err)
```
kevinconaway: nit, we should almost always use `require.NoError` so that tests fail fast.  `assert.NoError` records the failure but allows the test to continue which will probably panic

PR #306, opened 2019-05-08
Add conntrack support

leeavital:


ebpf/event_common.go
```diff
@@ -93,6 +93,22 @@ type ConnectionStats struct {
 	Type      ConnectionType      `json:"type"`
 	Family    ConnectionFamily    `json:"family"`
 	Direction ConnectionDirection `json:"direction"`
+	Conntrack *Conntrack          `json:"conntrack"`
+}
+
+// TODO: add to String()
+//easyjson:json
+type Conntrack struct {
+	OrigSrcIP string `json:"orig_src_ip"`
```
sunhay: Do we actually need to send back `Orig`? It should just be whatever we see in our source IP and dest IP, no?

checks/net.go
```diff
@@ -192,6 +193,23 @@ func formatDirection(d ebpf.ConnectionDirection) model.ConnectionDirection {
 	}
 }
 
+func formatConntrack(ct *ebpf.Conntrack) *model.Conntrack {
+	if ct == nil {
```
sunhay: Just wanted to confirm that the idea around this is that if there is no address translation, that we don't send anything to the backend. 

Is that accurate?

ebpf/event_common.go
```diff
@@ -93,6 +93,22 @@ type ConnectionStats struct {
 	Type      ConnectionType      `json:"type"`
 	Family    ConnectionFamily    `json:"family"`
 	Direction ConnectionDirection `json:"direction"`
+	Conntrack *Conntrack          `json:"conntrack"`
+}
+
+// TODO: add to String()
+//easyjson:json
+type Conntrack struct {
```
sunhay: Might be a nit, but I don't like the idea of naming this Conntrack. Can we call it something like `IPTranslation`, `AddrTranslation`, or `Translation`?

checks/net.go
leeavital: It's not, but it would be pretty easy to filter out the entries that don't have any translation.

ebpf/event_common.go
leeavital: Good point. If should always match the laddr.

ebpf/event_common.go
leeavital: I went with IPTranslation.

checks/net.go
sunhay: I think there's value in doing this to reduce message sizes. As a TODO for later :)

ebpf/config.go
```diff
@@ -37,6 +37,9 @@ type Config struct {
 
 	// BPFDebug enables bpf debug logs
 	BPFDebug bool
+
+	//  EnableConntrack enables probing conntrack for natural address translation via netlink
```
sunhay: nit: `natural` -> `network`

checks/net.go
leeavital: It was actually pretty easy to add. All the calls to `isNAT` are guarding against saving the non-NATd entries.

ebpf/tracer.go
```diff
@@ -80,13 +84,24 @@ func NewTracer(config *Config) (*Tracer, error) {
 
 	localAddresses := readLocalAddresses()
 
+	conntracker := netlink.NewNoOpConntracker()
+	if config.EnableConntrack {
+		c, err := netlink.NewConntracker()
+		if err != nil {
+			log.Errorf("could not initialize conntrack")
```
sunhay: I'd include the error here, as well as have a "continuing..." remark to highlight its non-fatal.

e.g. `log.Errorf("could not initialize conntrack: %s, continuing...", err)`

ebpf/tracer.go
```diff
@@ -80,13 +84,24 @@ func NewTracer(config *Config) (*Tracer, error) {
 
 	localAddresses := readLocalAddresses()
 
+	conntracker := netlink.NewNoOpConntracker()
+	if config.EnableConntrack {
+		c, err := netlink.NewConntracker()
+		if err != nil {
```
sunhay: It's a nit, but helps a lot with nested if-statements

```
if c, err := netlink.NewConntracker(); err != nil {
```

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+	"time"
+
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+
+	ct "github.com/florianl/go-conntrack"
+)
+
+// Conntracker is a wrapper around go-conntracker that keeps a record of all connections in user space
+type Conntracker interface {
+	GetConntrackEntryForConn(ip string, port uint16) *IPTranslation
```
sunhay: nit: this is redundant with the interface name. Just `GetTranslationForConn` should be fine.

ebpf/tracer.go
```diff
@@ -248,6 +267,12 @@ func (t *Tracer) getConnections(active []ConnectionStats) ([]ConnectionStats, ui
 	return active, latestTime, nil
 }
 
+var (
+	conntrackBuffer     [10000]ct.Conn
```
sunhay: Looks like these are unused? Perhaps old?

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+			log.Debugf("total registers: %d, ns/register: %f", ctr.stats.registers, float64(ctr.stats.registersTotalTime)/float64(ctr.stats.registers))
+		}
+		atomic.StoreInt64(&ctr.stats.gets, 0)
+		atomic.StoreInt64(&ctr.stats.getTimeTotal, 0)
+		atomic.StoreInt64(&ctr.stats.registers, 0)
+		atomic.StoreInt64(&ctr.stats.registersTotalTime, 0)
+	}
+}
+
+func isNAT(c ct.Conn) bool {
```
sunhay: Looks like we're calling these conversions multiple times `c.OrigSrcIP()`. Does it make sense to have this return more than just a `bool`?

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+
+	replSrcIP := ReplSrcIP(c)
+	replDstIP := ReplDstIP(c)
+
+	originSrcPort, _ := c.Uint16(ct.AttrOrigPortSrc)
+	originDstPort, _ := c.Uint16(ct.AttrOrigPortDst)
+	replSrcPort, _ := c.Uint16(ct.AttrReplPortSrc)
+	replDstPort, _ := c.Uint16(ct.AttrReplPortDst)
+
+	nat := !originSrcIP.Equal(replDstIP) ||
```
sunhay: `return` instead of doing an assignment

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+		originSrcPort != replDstPort ||
+		!originDstIP.Equal(replSrcIP) ||
+		originDstPort != replSrcPort
+
+	return nat
+}
+
+func ReplSrcIP(c ct.Conn) net.IP {
+	ipv4, ok := c[ct.AttrReplIPv4Src]
+	if ok {
```
sunhay:
```
if ipv4, ok := c[ct.AttrReplIPv4Src]; ok {
	return net.IPv4(ipv4[0], ipv4[1], ipv4[2], ipv4[3])
}
```

And similarly for v6 below.

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+
+	sessions, err = nfct.Dump(ct.Ct, ct.CtIPv6)
+	if err != nil {
+		// this is not fatal because we've already seeded with IPv4
+		log.Errorf("Failed to dump IPv6")
+	}
+	ctr.registerUnlocked(sessions)
+
+	go ctr.emitStats()
+	go ctr.compact()
```
sunhay: Can we just do both of these things in a single routine?

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+	return result
+}
+
+func (ctr *realConntracker) Close() {
+	ctr.statsTicker.Stop()
+	ctr.compactTicker.Stop()
+	ctr.nfct.Close()
+}
+
+func (ctr *realConntracker) registerUnlocked(sessions []ct.Conn) {
```
sunhay: Why is this called `registerUnlocked`?

sunhay: perhaps it can be renamed to `loadInitialState`?

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+	for _, c := range sessions {
+		if isNAT(c) {
+			ctr.state[formatKey(c)] = formatIPTranslation(c)
+		}
+	}
+}
+
+func (ctr *realConntracker) register(c ct.Conn) int {
```
sunhay: Oh this is cool. So this gets called by `go-conntrack` every time there's a new entry?

sunhay: A comment on the return value, as used by the other system would be useful here and on `unregister`

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+	ctr.state[formatKey(c)] = formatIPTranslation(c)
+
+	then := time.Now().UnixNano()
+	atomic.AddInt64(&ctr.stats.registers, 1)
+	atomic.AddInt64(&ctr.stats.registersTotalTime, then-now)
+
+	return 0
+}
+
+func (ctr *realConntracker) unregister(c ct.Conn) int {
```
sunhay: One thing I'm worried about here is with short-lived connections. Is there a race condition between close and when we actually unregister it?

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+type connKey struct {
+	ip   string
+	port uint16
+}
+
+type realConntracker struct {
+	nfct  *ct.Nfct
+	l     *sync.Mutex
```
sunhay: You can just make this `*sync.Mutex` and call `realConntracker.Lock()`

netlink/conntracker.go
```diff
@@ -0,0 +1,266 @@
...
+
+	ctr.l.Lock()
+	defer ctr.l.Unlock()
+
+	delete(ctr.state, formatKey(c))
+
+	return 0
+}
+
+func (ctr *realConntracker) compact() {
```
sunhay: Might be useful to have a comment that links to the golang issue that highlights constant map growth overtime (as added context on why we do this)

proto/agent.proto
```diff
@@ -371,7 +371,11 @@ message Connection {
 	uint32 lastRetransmits = 18;
 	ConnectionDirection direction = 19;
 
-	uint32 netNS = 20; // Network namespace
+  // Network namespace
+	uint32 netNS = 20;
+
+ // the conntrack entry associated with the connection. May be null on systems which don't support querying conntrack.
+  IPTranslation ipTranslation = 21;
```
sunhay: Whitespace looks wrong for all this.

sunhay:
Number of comments - this is looking really great though! :)

ebpf/tracer.go
```diff
@@ -225,6 +241,9 @@ func (t *Tracer) getConnections(active []ConnectionStats) ([]ConnectionStats, ui
 			if t.shouldSkipConnection(&conn) {
 				atomic.AddUint64(&t.skippedConns, 1)
 			} else {
+				// lookup conntrack in for active
+
+				conn.IPTranslation = t.conntracker.GetConntrackEntryForConn(conn.Source, conn.SPort)
```
sunhay: This only grabs it for connections currently in the BPF map (e.g. active connections). 

You'll want to do this above in `initPerfPolling` as well for closed connections.

netlink/conntracker.go
leeavital: They happen at different intervals

netlink/conntracker.go
leeavital: Looks like you realized. It's called registerUnlocked because the lock isn't held when it's called, so it's up to the caller (NewConntracker  to handle locking)

I like `loadInitialState` better though

netlink/conntracker.go
leeavital: Yep, until register returns 0,  or the handle is closed.  I'll document the return value.

netlink/conntracker.go
leeavital: Yes there is. I think need to have a "recently deleted" buffer that gets cleared by the tracer.

ebpf/tracer.go
leeavital: Agreed, this could be confusing. I think WARN is also more appropriate here.

ebpf/tracer.go
```diff
@@ -234,6 +248,8 @@ func (t *Tracer) getConnections(active []ConnectionStats) ([]ConnectionStats, ui
 			if t.shouldSkipConnection(&conn) {
 				atomic.AddUint64(&t.skippedConns, 1)
 			} else {
+				// lookup conntrack in for active
+				conn.IPTranslation = t.conntracker.GetTranslationForConn(conn.Source, conn.SPort)
```
sunhay: This still doesn't apply to short-lived connections.

I think you should move this call and `ClearShortLived()` to `GetActiveConnections()` after we've merged short-lived + active connections into a single list in 	

```
return &Connections{Conns: t.state.Connections(clientID, latestTime, latestConns)}, nil
```

netlink/conntracker.go
```diff
@@ -0,0 +1,305 @@
...
+	stats         struct {
+		gets               int64
+		getTimeTotal       int64
+		registers          int64
+		registersTotalTime int64
+	}
+}
+
+func NewConntracker() (Conntracker, error) {
+	nfct, err := ct.Open(&ct.Config{ReadTimeout: 10 * time.Millisecond, Logger: golog.New(os.Stdout, "go-conntrack", 0)})
```
sunhay: What effect is this logger going to have on our current logging library? Can we redirect them to it somehow?

netlink/conntracker.go
```diff
@@ -0,0 +1,305 @@
...
+		return nil, err
+	}
+
+	ctr := &realConntracker{
+		nfct:                nfct,
+		statsTicker:         time.NewTicker(time.Second * 10),
+		compactTicker:       time.NewTicker(time.Hour),
+		state:               make(map[connKey]*IPTranslation),
+		shortLivedBuffer:    make(map[connKey]*IPTranslation),
+		maxShortLivedBuffer: 10000,
```
sunhay: Let's make this configurable :)

ebpf/tracer.go
leeavital: I just pushed a change to handle short lived connections in initPerfPolling.  Thinking about it some more it's still racey. 

I don't follow your suggestion. GetActiveConnections is not doing any merging. All the merging happens in `state.go`

sunhay: Yeah, it gets merged via `t.state.Connections()` which is called in `GetActiveConnections()`. You could do it all there if you wanted.

What bit is still racey?

leeavital: > Yeah, it gets merged via t.state.Connections() which is called in GetActiveConnections(). You could do it all there if you wanted.

Are you suggesting adding the translations after calling `t.state.Connections()` ? I want to merge before (or inside `t.state.Connections()`) because the entries returned by `t.state.Connections` can be older which means entries need to stick around in `conntracker` longer.  Doing it inside  `t.state.connections() ` could work, but I'd like to avoid adding more code there if possible. 

> initPerfPolling is fine too though

> What bit is still racey?

There's no guarantee that the netlink callback to initialize the conntrack entry gets called before the `initPerfPolling` callback.

sunhay: >Are you suggesting adding the translations after calling t.state.Connections() ? I want to merge before (or inside t.state.Connections()) because the entries returned by t.state.Connections can be older which means entries need to stick around in conntracker longer.

Yeah the suggestion was to add it after calling `t.state.Connections()`. Aren't you storing all translations until this point anyways? I didn't see any code that clears it earlier than `ClearShortLived`.

> Doing it inside t.state.connections() could work, but I'd like to avoid adding more code there if possible.

Yeah, I'd prefer not to have anything more in here as well.

> There's no guarantee that the netlink callback to initialize the conntrack entry gets called before the initPerfPolling callback.

While it's not great memory wise, holding on to the translations for longer reduces this risk, since you're delaying the mapping of connection to it's translation. 

It would be nice if there was a way to not depend on that though.

sunhay: Either way, I think it's more than fine to continue what you're doing here with `initPerfPolling`, though you'll have to think of something re: that race condition.

netlink/conntracker.go
leeavital: I had enabled this early on for testing. In practice it wasn't helpful, so I've removed it for now.

PR #307, opened 2019-05-09
[network-tracer] Add more debug logs in error cases to the eBPF code

sfluor:
This adds more debug logs.

The logs lines are pretty short because there is a limit on the stack size:
```Looks like the BPF stack limit of 512 bytes is exceeded. Please move large on stack variables into BPF per-cpu array map.```


sunhay:
Does this actually fix the issue? I'm surprised it's adding up to more than 512 bytes.

sfluor:
Yep sorry the issue is gone but that was to justify the fact that logs lines are pretty short (using `src` instead of `source` for instance or `ERR` instead of `ERROR`)

PR #308, opened 2019-05-16
Use global net namespace for netlink if possible

leeavital:
When a netlink socket is created, messages for the current network
namespace are sent on the socket. This means that in a containernized
environment, we will by default get data for the containers network
namespace.

An easy workaround is to use `hostNetwork: true` when creating the
daemon. This isn't ideal.

This PR attempts to "find" the global namespace (which is reperesented
as a file descriptor) using the mounted host procfs and uses
go-conntrack's ability to temporarily change net namespace.

Requires the CAP_SYS_PTRACE capability, along with CAP_NET_ADMIN for
netlink.

### DIFF
--- a/netlink/conntracker.go
+++ b/netlink/conntracker.go
@@ -53,12 +53,14 @@ type realConntracker struct {
 }
 
 // NewConntracker creates a new conntracker with a short term buffer capped at the given size
-func NewConntracker(stbSize int) (Conntracker, error) {
+func NewConntracker(procRoot string, stbSize int) (Conntracker, error) {
 	if stbSize <= 0 {
 		return nil, fmt.Errorf("short term buffer size is less than 0")
 	}
 
-	nfct, err := ct.Open(&ct.Config{ReadTimeout: 10 * time.Millisecond})
+	netns := guessRootNetNSFd(procRoot)
+
+	nfct, err := ct.Open(&ct.Config{ReadTimeout: 10 * time.Millisecond, NetNS: netns})
 	if err != nil {
 		return nil, err
 	}
--- a/netlink/conntracker.go
+++ b/netlink/conntracker.go
@@ -58,7 +58,7 @@ func NewConntracker(procRoot string, stbSize int) (Conntracker, error) {
 		return nil, fmt.Errorf("short term buffer size is less than 0")
 	}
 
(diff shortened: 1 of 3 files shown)


netlink/net_ns.go
```diff
@@ -0,0 +1,28 @@
...
+	"os"
+
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+// guessRootNetNSFd guesses the file descriptor of the root net NS
+// and returns 0 in case of failure
+func guessRootNetNSFd(procRoot string) int {
+	for _, path := range []string{
+		procRoot + "/1/ns/net",
```
sunhay: `procRoot` will return either the `/host/proc` or `/proc`. Why do we need to try both?

netlink/net_ns.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"os"
+
+	"github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+// guessRootNetNSFd guesses the file descriptor of the root net NS
+// and returns 0 in case of failure
+func guessRootNetNSFd(procRoot string) int {
```
sunhay: `connectToRootNetNSFS` since we're not really guessing anymore?

netlink/conntracker.go
```diff
@@ -53,12 +53,14 @@ type realConntracker struct {
...
 		return nil, fmt.Errorf("short term buffer size is less than 0")
 	}
 
-	nfct, err := ct.Open(&ct.Config{ReadTimeout: 10 * time.Millisecond})
+	netns := guessRootNetNSFd(procRoot)
+
+	nfct, err := ct.Open(&ct.Config{ReadTimeout: 10 * time.Millisecond, NetNS: netns})
 	if err != nil {
 		return nil, err
 	}
```
sunhay: Since it only happens on startup, can we have a success INFO message?

``successfully created connection to netns fd: %s`, path`?

netlink/net_ns.go
leeavital: I suppose we don't. This is a holdover from when this was doing more of a guess

netlink/net_ns.go
leeavital: I went with `getGlobalNetNSFD`

PR #309, opened 2019-05-16
Improving storage of IP addresses

sunhay:
Use a single `Address` abstraction for both IPv4 and IPv6 which are backed by constant-sized byte arrays.


sunhay:
This is currently running in staging

ebpf/address.go
```diff
@@ -0,0 +1,95 @@
...
+	"net"
+
+	"github.com/mailru/easyjson/jwriter"
+)
+
+// Address is an IP abstraction that is family (v4/v6) agnostic
+type Address interface {
+	Bytes() []byte
+	String() string
+	MarshalEasyJSON(w *jwriter.Writer)
```
kevinconaway: nit: I believe you can embed the `easyjson.Marshaler` interface here for the same effect

ebpf/address.go
```diff
@@ -0,0 +1,95 @@
...
+	a[1] = byte(ip >> 8)
+	a[2] = byte(ip >> 16)
+	a[3] = byte(ip >> 24)
+	return a
+}
+
+// V4AddressFromString creates an Address using the string representation of an v4 IP
+func V4AddressFromString(ip string) Address {
+	var a v4Address
+	copy(a[:], []byte(net.ParseIP(ip))[12:16])
```
kevinconaway: You can avoid the cast here by just addressing the `ip` slice directly since it is a `[]byte`.

ebpf/event_test.go
```diff
@@ -79,72 +79,63 @@ func BenchmarkUniqueConnKeyByteBufferPacked(b *testing.B) {
...
 		p0 := uint64(c.Pid)<<32 | uint64(c.SPort)<<16 | uint64(c.DPort)
 		binary.Write(buf, binary.LittleEndian, p0)
-		buf.WriteString(c.Source)
+		buf.WriteString(c.SourceAddr().String())
 		// Family (8 bits) + Type (8 bits) = 16 bits
 		p1 := uint16(c.Family)<<8 | uint16(c.Type)
 		binary.Write(buf, binary.LittleEndian, p1)
-		buf.WriteString(c.Dest)
+		buf.WriteString(c.DestAddr().String())
 		buf.Bytes()
```
kevinconaway: Does this work?  I thought the compiler would optimize this away

ebpf/event_common.go
```diff
@@ -98,6 +99,16 @@ type ConnectionStats struct {
 	IPTranslation *netlink.IPTranslation `json:"conntrack"`
 }
 
+// SourceAddr returns the source address in the Address abstraction
+func (c ConnectionStats) SourceAddr() Address {
+	return c.Source.(Address)
```
leeavital: Should you handle nil here?

ebpf/address.go
leeavital: nit: I think `net.ParseIP(ip).To4()[0:4]` is more correct.     The docs for net.IP are vague about how an IPv4 is represented in 16 bytes, and `To4` explicitly returns 4 bytes.

ebpf/event_test.go
sunhay: I'm going to delete these benchmarks. They're really old and don't reflect the current state of the byte key generation

ebpf/event_common.go
sunhay: I had initially added it, but the amount of code needed to check it everywhere affected code readibility. This cast is safe to do everywhere except process-agent which does a string cast (which has comments on why).

In the next sprint or two, we'll rip out easy json and plugin protobuf (with the option to output as json as well for debugging) which will clean all this up

ebpf/address.go
sunhay: I purposely left it out to not include an `Unmarshal` function. I'm hoping we'll be able to rip all this out when we change to protobuf for process-agent <> network-tracer :)

kevinconaway: Understood.  FWIW, the `Marshaler` interface only contains the `MarshalXXX` method

PR #310, opened 2019-05-16
[conntrack]  Add stats for unregister

leeavital:


netlink/conntracker.go
```diff
@@ -45,10 +45,12 @@ type realConntracker struct {
...
 	stats         struct {
-		gets               int64
-		getTimeTotal       int64
-		registers          int64
-		registersTotalTime int64
+		gets                 int64
+		getTimeTotal         int64
+		registers            int64
+		registersTotalTime   int64
+		unregisters          int64
```
kevinconaway: Where are these incremented?

leeavital: They should be incremented in `unregister` but there was a typo

PR #311, opened 2019-05-16
network-tracer: Fix leaky stats objects that resulted in underflows

sfluor:
I took a dump of the state / added debug logs to the tracer and got:

```
--------- network-tracer logs --------------------

network-tracer.log:11:2019-05-16 14:36:57 UTC | NETWORK | ERROR | (ebpf/state.go:305 in updateConnWithStats) | underflow occured ! p:6687|src:172.21.58.160:9000|dst:172.21.3.190:22662|f:0|t:0

---------- ebpf logs -----------------------------

ebpf.logs:64486:         foo-6740  [003] d... 674427.956290: : closing conn sport:9000, dport:22662, daddr_l:-1107094100
ebpf.logs:97361:         foo-6696  [001] d... 674572.249423: : closing conn sport:9000, dport:22662, daddr_l:-1107094100

---------- state events dump ---------------------

CLOSE: {"pid":6687,"type":0,"family":0,"net_ns":4026531957,"source":"172.21.58.160","dest":"172.21.3.190","sport":9000,"dport":22662,"direction":1,"monotonic_sent_bytes":134,"last_sent_bytes":0,"monotonic_recv_bytes":4254,"last_recv_bytes":0,"monotonic_retransmits":0,"last_retransmits":0,"last_update_epoch":674411329457181}

CONNECTION: {"pid":6687,"type":0,"family":0,"net_ns":4026531957,"source":"172.21.58.160","dest":"172.21.3.190","sport":9000,"dport":22662,"direction":1,"monotonic_sent_bytes":0,"last_sent_bytes":0,"monotonic_recv_bytes":4254,"last_recv_bytes":0,"monotonic_retransmits":0,"last_retransmits":0,"last_update_epoch":674411316811701}

CLOSE: {"pid":6687,"type":0,"family":0,"net_ns":4026531957,"source":"172.21.58.160","dest":"172.21.3.190","sport":9000,"dport":22662,"direction":1,"monotonic_sent_bytes":177,"last_sent_bytes":0,"monotonic_recv_bytes":1283,"last_recv_bytes":0,"monotonic_retransmits":0,"last_retransmits":0,"last_update_epoch":674555618757552}
```

In this case the time between the `CONNECTION` event and the second `CLOSE` event is ~140s which is less than 2mn + 30s (our expiry limit for TCP connections + our polling interval), that's why the connection was not considered expired by the userspace code.

The main issue is that the second close event has less stats than the connection event (without a connection close between the two), and there is only 2 close events for this daddr, dport couple in the eBPF logs (that relates to the 2 we have in the state dump).

(I also had a script running that took a dump of the eBPF map every 15 seconds and it didn't have any connection for this (daddr, dport) couple)

There is likely something wrong with how we deal with close events but I don't know what's the issue.

I thought it was coming from close events we would drop here https://github.com/DataDog/datadog-process-agent/blob/master/ebpf/c/tracer-ebpf.c#L557 but I added a debug log there and it was not showing up so it does not seem to be the issue.

The goal of this PR is to at least fix the issue by clearing the old stats objects when we see a connection close event (except if the connection became active again). The issue could still appear if the interval between the CONNECTION and CLOSE event is smaller than our polling interval (30s).

It also fixes an other case where we could have underflows on closed connections becoming active again.

### DIFF
--- a/ebpf/state.go
+++ b/ebpf/state.go
@@ -141,7 +141,20 @@ func (ns *networkState) Connections(id string, latestTime uint64, latestConns []
 	// Update all connections with relevant up-to-date stats for client
 	conns := ns.mergeConnections(id, connsByKey)
 
-	// Flush closed connection map
+	// XXX: we should change the way we clean this map once
+	// https://github.com/golang/go/issues/20135 is solved
+	newStats := make(map[string]*stats, len(ns.clients[id].stats))
+	for key, st := range ns.clients[id].stats {
+		// Don't keep closed connections' stats
+		_, isClosed := ns.clients[id].closedConnections[key]
+		_, isActive := connsByKey[key]
+		if !isClosed || isActive {
+			newStats[key] = st
+		}
+	}
+	ns.clients[id].stats = newStats
+
+	// Flush closed connection map and stats
 	ns.clients[id].closedConnections = map[string]ConnectionStats{}
 
 	return conns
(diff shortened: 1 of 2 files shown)


ebpf/state.go
```diff
@@ -267,6 +276,28 @@ func (ns *networkState) mergeConnections(id string, active map[string]*Connectio
 	return conns
 }
 
+func (ns *networkState) updateConnWithStatsAndActiveConn(client *client, key string, active ConnectionStats, closed *ConnectionStats) {
```
sunhay: `updateConnWithStatWithActiveConn`?

Also add a comment on exactly why we need this function over `updateConnWithStats`

PR #313, opened 2019-05-20
Add retries when creating conntrack

leeavital:
I'm investigating a race condition where initializing conntrack
deadlocks. This is a hacky workaround that will get things working
until I find the root cause.

Also includes some logging adjustments that I made while debugging.

### DIFF
--- a/netlink/logging.go
+++ b/netlink/logging.go
@@ -1,0 +1,30 @@
+package netlink
+
+import (
+	"bufio"
+	"io"
+	"log"
+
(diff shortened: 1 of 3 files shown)


netlink/logging.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"bufio"
+	"io"
+	"log"
+
+	agentlog "github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+func getLogger() *log.Logger {
```
leeavital: this creates a go logger that delegates back the agent's logger package

PR #314, opened 2019-05-20
Use two conntrack sockets

leeavital:
go-conntrack has issues when we register two callbacks. To fix it, we
can use two conntrack connections.

I also included some logging improvements that were useful when debugging this.

### DIFF
--- a/netlink/conntracker.go
+++ b/netlink/conntracker.go
@@ -64,7 +64,7 @@ type realConntracker struct {
 }
 
 // NewConntracker creates a new conntracker with a short term buffer capped at the given size
-func NewConntracker(procRoot string, stbSize int) (Conntracker, error) {
+func NewConntracker(procRoot string, deleteBufferSize int) (Conntracker, error) {
 	var (
 		err         error
 		conntracker Conntracker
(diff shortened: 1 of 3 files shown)


netlink/conntracker.go
```diff
@@ -56,20 +64,47 @@ type realConntracker struct {
...
+
+	select {
+	case <-done:
+		return conntracker, err
+	case <-time.After(initializationTimeout):
+		return nil, fmt.Errorf("could not initialize conntrack after %s", initializationTimeout)
+	}
+}
+
+func newConntrackerOnce(procRoot string, stbSize int) (Conntracker, error) {
```
sunhay: While we're here, `stbSize` -> `deleteBufferSize`?

netlink/logging.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"bufio"
+	"io"
+	"log"
+
+	agentlog "github.com/DataDog/datadog-agent/pkg/util/log"
+)
+
+func getLogger() *log.Logger {
+	reader, writer := io.Pipe()
```
sunhay: Is this just the io pipe for this specific program?

leeavital: No. What this does is create an in-memory `Writer` that feeds into `reader`.  Go-conntrack takes a log.Logger which has a different interface than the agent logger.

sunhay: Should this return nil if we know the log level isn’t debug?

PR #317, opened 2019-05-28
Allow for hostname to be passed as encrypted environment variable

sunhay:
We were overwriting the hostname with whatever was passed as an environment variable. This should fix that.


remeh (approved):
LGTM

